Issue #18, opened 2022-11-28
[Chore] Setup CD pipeline for the project


nkhanh44:
## Why

- We need to set up the CD flow for this SwiftUI project so developers can distribute builds to `Firebase`.

## Acceptance Criteria

- We can run the CD pipeline on GitHubAction.
- We can distribute the build to `Firebase`.

PR #19, opened 2022-11-28
[#1] [Chore] Setup SwiftUI project

### LINKED ISSUE #1: [Chore] Setup SwiftUI project
## Why

Create a barebone SwiftUI project from Xcode.

## Acceptance Criteria

- SwiftUI project, the name can be "**Crypto Prices**", target >= iOS15.0
- Include Tests and UITests targets
- When running, the app shows a screen with the text "**Crypto Prices**" without any specifically required typography.

Thieurom:
- Close #1

## What happened 👀

Init the project with the `CryptoPrices` SwiftUI app

## Insight 📝

- Create a barebone SwiftUI project—named **CryptoPrices**—from Xcode.
- Include Tests and UITests targets (having no test for now)
- Target iOS>=15.0
- Put the source files in corresponding directories (`Sources` and `Resources`)
- Show a white screen with the text "**Crypto Prices**" on the launch
- Update `gitignore` by removing unrelated platforms stuff (`Node`, `Ruby`)

## Proof Of Work 📹

Screenshot_2022-11-28_at_15_09_06

Simulator Screen Shot - iPhone 14 Pro - 2022-11-28 at 15 08 57

### DIFF
--- a/CryptoPrices/Sources/ContentView.swift
+++ b/CryptoPrices/Sources/ContentView.swift
@@ -1,0 +1,27 @@
+//
+//  ContentView.swift
+//  CryptoPrices
+//
+//  Created by Doan Thieu on 28/11/2022.
+//
+
+import SwiftUI
+
+struct ContentView: View {
+
+    var body: some View {
+        VStack {
+            Text("Crypto Prices")
+                .font(.system(size: 28.0))
+                .fontWeight(.medium)
+        }
+        .padding()
+    }
+}
+
+struct ContentView_Previews: PreviewProvider {
+
+    static var previews: some View {
+        ContentView()
+    }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -1,36 +1,5 @@
-*.gem
-*.rbc
-/.config
-/.idea
-/coverage/
-/InstalledFiles
-/node_modules
-/pkg/
-/spec/reports/
-/spec/examples.txt
-/test/tmp/
-/test/version_tmp/
-/tmp/
+# Mac
+.DS_Store
 
-# Used by dotenv library to load environment variables.
-# .env
-
-## Documentation cache and generated files:
-/.yardoc/
-/_yardoc/
-/doc/
-/rdoc/
-
-## Environment normalization:
-/.bundle/
-/vendor/bundle
-/lib/bundler/man/
-
-# for a library or gem, you might want to ignore these files since the code is
-# intended to run in multiple environments; otherwise, check them in:
-# Gemfile.lock
-# .ruby-version
-# .ruby-gemset
-
-# unless supporting rvm < 1.11.0 or doing something fancy, ignore this:
-.rvmrc
+# Xcode
+xcuserdata/
--- a/CryptoPrices.xcodeproj/project.pbxproj
+++ b/CryptoPrices.xcodeproj/project.pbxproj
@@ -1,0 +1,622 @@
+// !$*UTF8*$!
+{
+	archiveVersion = 1;
+	classes = {
+	};
+	objectVersion = 56;
+	objects = {
+
+/* Begin PBXBuildFile section */
+		4D0A1DA029349D660038624D /* CryptoPricesApp.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4D0A1D9F29349D660038624D /* CryptoPricesApp.swift */; };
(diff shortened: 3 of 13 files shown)


minhnimble (approved):
I do have a few questions:
- Just curious that do we need to setup some common SPM dependencies in this chore also? 
- Do we need to setup and use any 3rd libraries for writing tests (especially unit tests) or just use `XCTest` for this project?
- It looks like there is no development team assigned to the project at the moment and thus we might not be able to build this application on a real device. Should we set the development team to `Nimble` also? (You guys might need to request access to our Nimble's Apple Developer Account 🙏)
  Screenshot 2022-11-29 at 10 13 32
  Screenshot 2022-11-29 at 10 13 56

Apart from the above concerns, I built and ran the project successfully on my local, nice work 🚀 Thus, the rest LGTM 🎉

Thieurom:
@minhnimble Thanks for the feedback 🙏 I'd like to share my idea on your concerns as follows:
- This PR includes initializing a very _barebones_ SwiftUI project only, no more, no less 🙏
  For the dependencies we won't add them all at once beforehand, we'll add them gradually when we actually need them for implementation. For e.g. when we need to create Skeleton loading we add a library for that. Or when we implement the prices charts we add another one at that time.
- For writing unit tests, when we add the first unit tests (maybe on the first backend task) I think we can use the `Quick/Nimble` framework. It supports `async/await` starting from `6.0.0` and its syntax is really short and readable (to me it's more pleasant than writing with `RxSwift` or `Combine`) This is the example on [their docs](https://github.com/Quick/Quick/blob/main/Documentation/en-us/AsyncAwait.md):
  ```swift
  describe("my awesome tests") {
      beforeEach {
          await someAsyncSetupFunction()
      }

      afterEach {
          await teardownButAsync()
      }
    
      aroundEach { runExample in
          await runExample()
      }

      it("can run tests asynchronously!") {
          let result = await methodBeingTested()
          expect(result).to(equal(expectedValue))
      }
  }
  ```
  Writing with `XCTest` is also as simple and readable as that because with `async/await` we don't have to use `expectation` anymore:
  ```swift
  func testAMethod() async {
      let result = await methodBeingTested()
      XCTAssertEqual(result, expectedValue)
  } 
  ```
  We can have further discussion on our channel 🙏 
- For building on real devices:
  - Debug build: we can let Xcode manages to sign. We can choose to use our personal account or Nimble account.
  - Release build for distributing on the Firebase: We can also choose between two options: use Ad-hoc (add our and tester UUIDs with Nimble account to create certs and profiles) or just use email invitation.
  
  However, I'd prefer to address these later when we do [the setup CD task](https://github.com/nimblehq/nimble-crypto-ios/issues/18). What do you think?

minhnimble:
@Thieurom all the points are making sense for me, thanks for the detailed feedback 🙏 

- Regarding the test library, then I would prefer to use the `Quick/Nimble` framework thanks to its syntax sugar and the familiarity of the team with it. And I guess we can start adding this library when we start writing the first test, not here in this chore PR 👌
- Debug build: I would prefer to have a shared account which can be Nimble account by default so that we can commit to the codebase repository with this chore task. Choosing to use our personal account will require us to only modify locally which is kinda cumbersome, to be honest. 😅
- Release build for distributing on the Firebase: Yeah this is definitely something we should should setup in the CD task, you don't have to worry about it here since it is not my main concern for this PR 💪

CryptoPrices/Sources/ContentView.swift
```diff
@@ -0,0 +1,27 @@
...
+//  Created by Doan Thieu on 28/11/2022.
+//
+
+import SwiftUI
+
+struct ContentView: View {
+
+    var body: some View {
+        VStack {
+            Text("Crypto Prices")
```
nkhanh44, line 14: Do you think we should add `TODO: -` to remind us to remove it later?

nkhanh44 (approved):
Nice work 💯 .

CryptoPrices/Sources/ContentView.swift
Thieurom: Normally the `TODO` means we can put it off, for now, and we can come back to address it later. But I think this is not necessary since this is kind of __must do__.
Also, we can look at it as the _current UI design_ from the designer, not the placeholder one 😅

PR #27, opened 2022-12-05
[#24] [Chore] Create the Home module

### LINKED ISSUE #24: [Chore] Create the Home module as a local Swift package
## Why

We need to create a module so that the Home screen's source files (`View`, `ViewModel`, ...) can be put on it.
The module will be named __Home__ and created as a local Swift package.

## Who Benefits?

Developers can start developing for the Home screen.

Thieurom:
- Close #24

## What happened 👀

We need to create a module so that the Home screen's source files (View, `ViewModel, ...) can be put on it.
The module will be named Home and created as a local Swift package.

## Insight 📝

- Create `Styleguide` local package for typography (fonts), colors, assets, common custom views, ...
- Add a `Preview` helper to preview SwiftUI views in both Light and Dark modes.
- Create `Home`  local package for all source files relating to the Home screen.
- Link `Home` to the main app

## Proof Of Work 📹

Screenshot_2022-12-05_at_11_50_46

|Light mode preview|Dark mode preview|
|---|---|
|Screenshot 2022-12-05 at 11 38 43|Screenshot 2022-12-05 at 11 48 44|


CryptoPrices/Sources/Styleguide/Sources/Styleguide/SwiftUIHelpers/Preview.swift
```diff
@@ -0,0 +1,34 @@
...
+                .previewDevice("iPhone 14 Pro")
+                .previewDisplayName("Pro - Light mode")
+
+            content
+                .environment(\.colorScheme, .dark)
+                .preferredColorScheme(.dark)
+                .previewDevice("iPhone 14 Pro")
+                .previewDisplayName("Pro - Dark mode")
+        }
+    }
```
minhnimble, line 33: Instead of disable `swiftlint` here with `swiftlint:disable type_contents_order`, I think we can just simply reorder the function here a bit as follow:
```suggestion
// swiftlint:disable type_contents_order

import SwiftUI

public struct Preview<Content: View>: View {

    let content: Content

    public var body: some View {
        Group {
            content
                .environment(\.colorScheme, .light)
                .preferredColorScheme(.light)
                .previewDevice("iPhone 14 Pro")
                .previewDisplayName("Pro - Light mode")

            content
                .environment(\.colorScheme, .dark)
                .preferredColorScheme(.dark)
                .previewDevice("iPhone 14 Pro")
                .previewDisplayName("Pro - Dark mode")
        }
    }
    
      public init(@ViewBuilder _ content: () -> Content) {
        self.content = content()
    }
```

CryptoPrices/Sources/Styleguide/README.md
```diff
@@ -0,0 +1,3 @@
+# Styleguide
+
+Typography, colors, assets, common views, and other styling-related stuff.
```
minhnimble, line 3: Since it looks like the `Styleguide` module only supports the view's related stuff, how about shared non-view related stuff like public booleans, other constants, device settings, translations, etc? It seems that we will still need to create another shared module, right? So, why don't we turn the `Styleguide` into `Shared` instead for future extendability?

CryptoPrices/Sources/Styleguide/Sources/Styleguide/SwiftUIHelpers/Preview.swift
Thieurom: Thanks, @minhnimble 🙏 

I think everyone (including me at first) just simply reorder a little bit to satisfy `swiftlint` here. But the more I think about it, the more it annoys me 😅

That's because the SwiftUI `View` is special and different from other types, like data structures or view models, ... The SwiftUI `View` merely is a description of a view, and almost its code lives in the `body` computed property. And this `body` computed property is quite long depending on how complex the layout is. It isn't simply with one or a few lines of codes like `var httpMethod: String { "GET" }` as we normally see. As a result, the initializer, which is the very first API that we use, the one that tells us how we initialize, construct, and create a new instance easily gets lost at the very bottom of the view definition.

I did some research to conditionally opt out of this `type_contents_order` rule specifically for SwiftUI views but got no luck. This issue has been even raised on the `swiftlint` repo [here](https://github.com/realm/SwiftLint/issues/3548) but no progress so far.

But I think for now I can go with the reordering approach and let all of us experience this matter and have a further discussion along the way 🙏

minhnimble: @Thieurom the `body` var is indeed a special and potentially long variable in SwiftUI for sure. At the same time, it is all coming down to the perspective of the developers and in this case, it can be seen that we have a slightly different view at the moment as I am not really having an issue with placing the init function a bit below 😅. For me, in our previous ViewModels that have huge logic with many states, we would need to create a lot of variables before the `init` call, right? At the same time, grouping variables and functions correspondingly allow much easy searching and organization, especially when we combine it with the `MARK` usage. Then we can use Xcode's feature to find what we want and later on where to add new vars/functions with ease 🙏:

Screenshot 2022-12-06 at 10 08 47

-> Anyhow, if you really want to commit to placing the init function a bit earlier at the top level and mix them in between, setting `type_contents_order` rule on the file is still not a great approach because it is going to be a recurring problem. We will have to add this rule on every file that needs the `init` function, which might also introduce more potential issues that some new devs might forget/ be unaware of the requirement here. A better way could be disabling the rule directly on the `.swiftlint.yml` file for having a global effect 💪

CryptoPrices/Sources/Styleguide/README.md
Thieurom: Thanks, @minhnimble for the concerns! I'd like to share my idea about how we're going to modularize our application 🙏 
- As its name implies and its description, the `Styleguide` module contains only the view's related stuff. That's because in order to leverage modulization, we have to think carefully about which stuff should go to one or another module. For e.g in case, we need to expose common constants and settings, we should group them into a separate module like `Constants` or `Settings`. It has some great benefits like:
  - Avoid a large, giant shared module. We should try to further divide and package stuff into smaller modules.
  - Reason about how to use from the perspective of the module consumer (importer). Let's say we combine all view and settings stuff into a single module `SharedModule`. So a view-related module like `Home` in order to use just some fonts, colors, and common views has to import all include the data stuff they don't want. On the other hand, it's a little bit strange that a data-related module like `Repository` in order to use some settings, constants like `apiURL` imports a module with view stuff.
  - Easily organize, develop, and maintain.
  - Good if not great build time. If we package (group) related stuff to different, separate modules, the compiler doesn't have to build a single, large module which may take a long time. It can parallelly build multiple modules at the same time. And speaking of code changes, the change on this module won't cause to rebuild of unrelated modules. For e.g. the `Settings` module which contains some constants that rarely change, we almost _set and forget_. But the `Styleguide` module may change more frequently for some reasons like we add a new font style, or changing the alpha of the existing color, ... If we put them all into a single module, any small change like adding new font will cause the whole big module rebuild, include the Swift files that declare some constants.

- About the naming of the modules, I think we should name them based on ___what__ they provide and expose_ rather than ___how__ they are being used_. More clearly is that we shouldn't bother with the `shared` word even if they're being _shared_ in the sense that one or more other parts will import and use them. For e.g. instead of `SharedViews` we should name it `Styleguide`, `DesignSystem`, or `CustomViews`. Instead of `SharedModels`, we can go with `DomainModels` or `Entities` 🙏

minhnimble: @Thieurom I think you had valid reasons there, great points 👍 More small modules even for the shared category there should be a great direction as long as we manage the potential duplication well. Let's stick with this approach then. I think we should note this somewhere for later usage 😁

CryptoPrices/Sources/Styleguide/Sources/Styleguide/SwiftUIHelpers/Preview.swift
Thieurom: @minhnimble Maybe I'm a little bit unclear here 🙏  When I say `gets lost` it's not about how hard to navigate to (by means of `MARK` like you said, which I'm aware of and definitely agree to apply). What I just mean is the way I look at the code, it's more about aesthetics, easy in the eye I guess 😅 . And as I [provided](https://github.com/realm/SwiftLint/issues/3548), it's not only me who has the same experience, and the same feeling but there're some other guys too. So I think it's maybe just a matter of preference, which is subjective, so feel free to ignore it 😄
> A better way could be disabling the rule directly on the .swiftlint.yml file for having a global effect.

This rule should be encouraged to apply across the app, that's why I was trying to conditionally opt out specifically for SwiftUI views only, not globally but `swiftlint` does not support it now. Until then we don't any other way than to manually disable file by file. One quick way I can think of right now is using a code generator (like `sourcery` to prepend that disabling rule at the top of every SwiftUI file). But maybe I overthink it, and I'm totally fine with the reordering approach to avoid making things complicated.

minhnimble (approved):

nkhanh44 (approved):
LGTM 👍

PR #30, opened 2022-12-08
[#29] [Chore] Create the MyCoin module as a local Swift package

### LINKED ISSUE #29: [Chore] Create the MyCoin module as a local Swift package
## Why

We need to create a module so that the _MyCoin_ screen's source files (View, ViewModel, ...) can be put on it.
The module will be named _MyCoin_ and created as a local Swift package.

## Who Benefits?

Developers can start developing for the _MyCoin_ screen.

minhnimble:
- Close #29

## What happened 👀

- Create `MyCoin` local package for all source files relating to the MyCoin screen.
- Link the `MyCoin` SPM package to the main application for later usage + navigation.

## Insight 📝

Right now, since the scale of the project is small and we have only 2 main screens, the modular package is introduced accordingly per screen. For bigger projects, we should create a package per feature that could contain multiple screens.

## Proof Of Work 📹

- The new `MyCoin` local package is available in the project:
 
Screenshot 2022-12-08 at 10 32 29

- Hard code to show the current `MyCoin` view:

|Light mode preview|Dark mode preview|
|--------------------|-------------------|
|||

### DIFF
--- a/CryptoPrices.xcodeproj/project.pbxproj
+++ b/CryptoPrices.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		249752BD2941901E003C6238 /* MyCoin in Frameworks */ = {isa = PBXBuildFile; productRef = 249752BC2941901E003C6238 /* MyCoin */; };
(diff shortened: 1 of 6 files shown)


nkhanh44 (approved):
LGTM

Thieurom (approved):
LGTM 🚀

PR #33, opened 2022-12-09
[#6] [Backend] As a user, I can see My Coins on the Home screen

### LINKED ISSUE #6: [Backend] As a user, I can see My Coins on the Home screen
## Why

- The user should be able to see their coins on the Home screen.

## Acceptance Criteria

- When the user navigates to the Home screen, the following API is called: `GET /coins/markets`.

Parameters:
```
    Currency: vs_currency=usd
    Items per page: per_page=10
    Items order: order=market_cap_desc
    Page: page=1
    Price changes percentage: price_change_percentage=24h
```
- If the API returns success, the user will see the coins items (handled in integrate story #5).
- If the API returns error, the user will see the error state using default alert with title is `Error`, message's content is from API response  and the confirm button's title is `Ok`.

``` JSON
{
  "error": "Could not find coin with the given id"
}
```
## Design
- [Figma design](https://www.figma.com/file/Iy9eg3ym29pkKGsD5X48O7/Crypto-App-(Community)?node-id=3%3A181&t=W7iFTb0v4cHHYenL-4).

Thieurom:
- Close #6

## What happened 👀

- Create `TestHelpers` package to support unit testing
- `Data` package:
  - Divide into smaller modules: `NetworkCore`, `NetworkExtension`, and `Repositories` since `Data` is a too vague name, and we mixed interface and implementation.
  - Add `DataTestHelpers` for dummy and mock data
  - Implement fetching my coins for `CoinRepository` + unit tests
- `Domain` package:
  - Add `DomainTestHelpers` for dummy and mock data
  - Implement `MyCoinsUseCase` + unit tests
- Update linting rule:
  - Remove [`nimble_operator`](https://realm.github.io/SwiftLint/nimble_operator.html) for unit testing with `Quick/Nimble`
  - Update [`closure_body_length`](https://realm.github.io/SwiftLint/closure_body_length.html) from the default of `30` to `40`

## Insight 📝

- We created `TestHelpers` package to support unit testing. Other modules can import it to write unit tests more easily. This package doesn't depend on anything.
- Each package can have its own test helpers module to support unit testing its implementation. In this PR we apply to `Data` and `Domain`
- We might need to use a code generator (e.g. `Sourcery`) for generating mock types.

## Proof Of Work 📹

- Later we need to update `fastlane` to test local packages (together with the app)

Screenshot 2022-12-12 at 10 48 30
Screenshot 2022-12-12 at 10 48 56


CryptoPrices/Sources/Data/Sources/NetworkCore/APICoin.swift
```diff
@@ -0,0 +1,19 @@
+//
+//  APICoin.swift
+//  Data
+//
+//  Created by Doan Thieu on 01/12/2022.
+//
+
+import Foundation
+
+public struct APICoin: Decodable, Equatable {
```
nkhanh44, line 10: IMO in NetworkCore, we should define the core of configuration, API error ... or stuff like that, and APICoin is a model. Because of that, for clarifying, what do you think we should move model APICoin to its Model folder? 💭

CryptoPrices/Sources/Data/Sources/DataTestHelpers/Mock/MockCoinAPI.swift
```diff
@@ -0,0 +1,38 @@
...
+public class MockCoinAPI: CoinAPIProtocol {
+
+    public var myCoinsReturnValue: Result<[APICoin], Error>?
+
+    public init(myCoinsReturnValue: Result<[APICoin], Error>? = nil) {
+        self.myCoinsReturnValue = myCoinsReturnValue
+    }
+
+    public func myCoins() async throws -> [APICoin] {
+        guard let myCoinsReturnValue = myCoinsReturnValue else {
```
nkhanh44, line 21: FYI, we could shorten it like this in swift 5.7 :D:
```suggestion
        guard let myCoinsReturnValuee else {
```

CryptoPrices/Sources/Domain/Sources/DomainTestHelpers/MockCoinRepository.swift
```diff
@@ -0,0 +1,39 @@
...
+public class MockCoinRepository: CoinRepositoryProtocol {
+
+    public var myCoinsReturnValue: Result<[Coin], Error>?
+
+    public init(myCoinsReturnValue: Result<[Coin], Error>? = nil) {
+        self.myCoinsReturnValue = myCoinsReturnValue
+    }
+
+    public func myCoins() async throws -> [Coin] {
+        guard let myCoinsReturnValue = myCoinsReturnValue else {
```
nkhanh44, line 22: same here
```suggestion
        guard let myCoinsReturnValue else {
```

CryptoPrices/Sources/Domain/Tests/UseCasesTests/MyCoinsUseCase/MyCoinsUseCaseTests.swift
```diff
@@ -1,11 +1,54 @@
+import DomainTestHelpers
 import Nimble
 import Quick
+import TestHelpers
 @testable import UseCases
 
 final class MyCoinsUseCaseTests: QuickSpec {
```
nkhanh44, line 7: This is optional. I think using the suffix `Spec` is suitable in this case. Because when using lib Nimble & Quick, we used the suffix `Spec` for consistency as [a convention](https://nimblehq.co/compass/development/code-conventions/swift/quick-nimble/#convention-1). I think it makes sense for the suffix `Tests` when we conform XCTestCase :D.

I also think it can be flexible and understood by developers with the suffix "Tests". 😄

nkhanh44 (commented):
Just a few comments. The work looks great to me 😄 .

CryptoPrices/Sources/Data/Sources/NetworkExtension/MyCoinsParameters.swift
```diff
@@ -0,0 +1,15 @@
+//
+//  MyCoinsParameters.swift
+//  Data
+//
+//  Created by Doan Thieu on 09/12/2022.
+//
+
+public struct MyCoinsParameters: Parameterable {
```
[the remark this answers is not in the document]
Thieurom: The `MyCoinsParameters` (and `Parameterable` for that matter) is actually a part of the implementation detail (`NetworkExtension`), not a part of the core/interface (`NetworkCore`). For e.g, we might have another implementation of `CoinAPIProtocol` which doesn't use any of `MyCoinsParameters` (or `Pilot`, `CoinRoute`, ...) at all. (It might just pass the parameters `perPage` or `page` directly if you got the point 🙏 ) That's why `MyCoinsParameters` belongs to `NetworkExtension`, not a separate module.

CryptoPrices/Sources/Data/Sources/NetworkCore/APICoin.swift
Thieurom: I think the only _things (objects)_ that belong to _Model_ are the ones we put in the `Entities` module of the `Domain` package. Because this is the center of the app, to make the app according to the required business logic many modules would depend on the `Entities` module, from `Data`, `Home`, and `MyCoin` to the app itself.

Speaking of `APICoin`, it is of course the _model_ in the sense that is a concrete type, but it is never used outside the context of the network layer (hence the `API` prefix). The only modules that depend on it are the ones that implement the network layer (`NetworkExtension` in this case).

You can see that as there's no Repository, UseCase and the app ever use this `APICoin`. They all work on the `Coin` abstraction defined in the `Entities`. They're free to define their own _model_ that conforms to the `Coin` interface. For e.g, the `Home` module might have `MyCoinUIModel` doing that. But in any case, the `APICoin` is never propagated up to a layer that's higher than the network.

That's why the `APICoin` belongs to the `NetworkCore`, and together with the `CoinAPIProtocol` to provide a blueprint, a specification of the network layer: `func myCoins() async throws -> [APICoin]`, etc, ...

CryptoPrices/Sources/Data/Sources/DataTestHelpers/Mock/MockCoinAPI.swift
Thieurom: Actually, I intend to make this app compatible with Swift 5.5 (Xcode 13.2.1 if I remember correctly). You can see in all the local package's manifest (`Package.swift` file), the very first lines are:
```
// swift-tools-version: 5.5
// The swift-tools-version declares the minimum version of Swift required to build this package.
```

So while this is nice and improved syntax, it won't compile with Xcode with versions of `13.2.1` ~ < `14.0.0`.
Later if we decide to lift up the whole project to the latest Swift version, we can have a dedicated to do so. It's because it's not simple as just using the newer Xcode. There are chances that third-party libraries won't be compatible 🙏

CryptoPrices/Sources/Data/Sources/NetworkCore/APICoin.swift
nkhanh44: Thank you for the detailed explanation 🙇 . It's clear to me now.

nkhanh44 (approved):
💯 🚀 🌔

CryptoPrices/Sources/Domain/Sources/DomainTestHelpers/TestCoin.swift
```diff
@@ -0,0 +1,34 @@
...
+//  TestCoin.swift
+//  Domain
+//
+//  Created by Doan Thieu on 12/12/2022.
+//
+
+import Entities
+import Foundation
+
+public struct TestCoin: Coin, Equatable {
```
minhnimble, line 11: `MockCoin` might be a better name here, what do you think? 🤔
```suggestion
public struct MockCoin: Coin, Equatable {
```

CryptoPrices/Sources/Data/Tests/RepositoriesTests/CoinRepository/CoinRepositorySpec.swift
```diff
@@ -0,0 +1,55 @@
...
+import TestHelpers
+@testable import Repositories
+
+final class CoinRepositorySpec: QuickSpec {
+
+    override func spec() {
+        var coinRepository: CoinRepository!
+        var coinAPI: MockCoinAPI!
+
+        describe("The CoinRepository") {
```
minhnimble, line 14: The capitalization of the word `The` might be unnecessary here 🙏
```suggestion
        describe("the CoinRepository") {
```

CryptoPrices/Sources/Data/Sources/NetworkExtension/CoinRoute.swift
```diff
@@ -11,19 +11,35 @@ import PilotType
...
-    public var parameters: Parameters? { nil }
+
+    public var parameters: Parameters? {
+        switch self {
+        case let .myCoins(parameters): return parameters.encoded()
+        default:
+            // TODO: Update when implement
+            return nil
+        }
+    }
```
minhnimble, line 42: Small code refactor to be consistent with `path` 🙏 
```suggestion
    public var parameters: Parameters? {
        switch self {
        case let .myCoins(parameters): return parameters.encoded()
           // TODO: Update when implement
        default: return nil
        }
    }
```

CryptoPrices/Sources/Data/Sources/DataTestHelpers/Mock/MockCoinAPI.swift
```diff
@@ -0,0 +1,38 @@
...
+//  Data
+//
+//  Created by Doan Thieu on 09/12/2022.
+//
+
+import NetworkCore
+
+// TODO: - Set up code generator for generating mock types (e.g. Sourcery)
+
+public class MockCoinAPI: CoinAPIProtocol {
```
minhnimble, line 12: Why don't we apply `Sourcery` right away? This code is going to be removed once we have `Sourcery` in place, which is a bit of wasted effort here to imo 😅

CryptoPrices/Sources/Domain/Sources/DomainTestHelpers/TestCoin.swift
Thieurom: Yeah, the `Mock` prefix would make it consistent with other types, like `MockCoinRepository` in the same module 👍

CryptoPrices/Sources/Data/Sources/NetworkExtension/CoinRoute.swift
Thieurom: Nice catch with attention to detail 😄 Will update soon!

CryptoPrices/Sources/Data/Sources/DataTestHelpers/Mock/MockCoinAPI.swift
Thieurom: The reason is that this PR is quite large, so I'd like to focus on how we would implement the first backend task (with the very first unit tests to be written as well) 🙏 

Speaking of a code generator for stub and mock, `Sourcery` is great and I might use it. But I'm still looking at other alternatives like [Mockingbird](https://github.com/birdrides/mockingbird), [Cuckoo](https://github.com/Brightify/Cuckoo), ... The idea is that not only we want to use the popular, widely-used and familiar tools but also we want to broaden the options, and solutions so we ourselves can have more skills under our belt 🙏 
 
Also, the number of mocks I have to write by hand is quite small (`MockAPI` and `MockCoinRepository`), so I just went ahead and added them first with considering replacing them later (hence the `TODO` marks).

CryptoPrices/Sources/Domain/Sources/DomainTestHelpers/TestCoin.swift
Thieurom: Updated with [#062134e](https://github.com/nimblehq/nimble-crypto-ios/pull/33/commits/062134eefb9111754b288ed6614af17451aa4eed)!

CryptoPrices/Sources/Data/Tests/RepositoriesTests/CoinRepository/CoinRepositorySpec.swift
Thieurom: Updated with [#062134e](https://github.com/nimblehq/nimble-crypto-ios/pull/33/commits/062134eefb9111754b288ed6614af17451aa4eed)!

CryptoPrices/Sources/Data/Sources/NetworkExtension/CoinRoute.swift
Thieurom: Updated with [#062134e](https://github.com/nimblehq/nimble-crypto-ios/pull/33/commits/062134eefb9111754b288ed6614af17451aa4eed)!

CryptoPrices/Sources/Data/Sources/DataTestHelpers/Mock/MockCoinAPI.swift
minhnimble: Noted, thank you for your though process there, I really like the idea that this project can be a good playground for us to experiment new tools 👍 Let's keep it as-is and hold the Sourcery integration for now 💪, and let's have a discussion to decide which mocking tool we would like to use for the project later then 🚀

minhnimble (approved):
LGTM 🚀

PR #40, opened 2022-12-13
[#35] [Chore] Set up code generator for strings, fonts, colors, images and other types of resources

### LINKED ISSUE #35: [Chore] Set up code generator for strings, fonts, colors, images and other types of resources
## Why

We have to set up a code generator for strings, fonts, colors, images, and other types of resources so we can access them by Swift code, not by string-based one.
- For this purpose, we're going to use [`swiftgen`](https://github.com/SwiftGen/SwiftGen) library.
- Add a command to `Makefile` so that when running it'll (re)generate Swift code for all packages that embed the resources (`Styleguide`, `Home`, `MyCoin`, ...)

## Who Benefits?

Developers who need to access the resources by Swift code while developing the app and feature modules.

Thieurom:
- Close #35

## What happened 👀

- Set up the installation and running `swiftgen` in `Makefile`
- Generate Swift code for:
  -  assets (colors and images) and fonts for the `Styleguide` package. This includes moving all assets from the app to the `Styleguide` and converting all predefined colors from code (`Color+Extension.swift`) to the asset catalog.  These are public so that the app and other modules can import them.
  - localized strings for the `Home` package. These are internally used in the `Home`.
- Remove now-become-unnecessary code (mainly because we now have the equivalent generated by `swiftgen`)
- Update views with newly-created resources API.
- Make linting code consistent all places with `make lint` command:
  - Xcode build phase
    Screenshot 2022-12-15 at 10 41 58

  - Commandline
    Screenshot 2022-12-15 at 10 46 09

  - CI
    Screenshot 2022-12-15 at 10 44 07

## Insight 📝

- `swiftgen` is quite powerful and easy to config.
- GitHub action sometimes makes me frustrated 😅 

## Proof Of Work 📹

The app runs with the same data and appearance as before (colors, images, fonts, texts) 😄
(Makes me think of adding snapshot testing to our project 🤔 )
| Light mode | Dark mode|
|---|---|
|Simulator Screen Shot - iPhone 14 Pro - 2022-12-14 at 11 15 43|Simulator Screen Shot - iPhone 14 Pro - 2022-12-14 at 11 15 54|


minhnimble:
Can we add some instructions usage on the new tool - `swiftgen` here? 🙏 This is because when I checkout the source code and add a new localized string, the `Strings` file doesn't get updated automatically with the new contents. 😅

CryptoPrices/Sources/Home/Sources/Home/Localization/Strings.swift
```diff
@@ -1,20 +1,61 @@
...
-//
-//  Created by Minh Pham on 06/12/2022.
-//
+// swiftlint:disable all
+// Generated using SwiftGen — https://github.com/SwiftGen/SwiftGen
 
 import Foundation
 
-public enum Strings {
+// swiftlint:disable superfluous_disable_command file_length implicit_return prefer_self_in_static_references
```
nkhanh44, line 6: Are all the `swiftlint: disable` generated by SwiftGen? As I see, we have `swiftlint:disable all` at the top file, I think the rest `swiftlint: disable`s are not needed.

CryptoPrices/Sources/Home/Sources/Home/WalletStatisticSection.swift
```diff
@@ -53,26 +58,26 @@ private extension WalletStatisticSection {
     var todayProfit: some View {
         HStack {
             VStack(alignment: .leading, spacing: 8.0) {
-                Text(Strings.Home.todayProfitTitle)
-                    .foregroundColor(.lightSilver)
-                    .font(.Inter.medium.font(for: .callout))
+                Text(Strings.Home.TodayProfit.title)
+                    .foregroundColor(Assets.Colors.lightSilver.swiftUIColor)
```
nkhanh44, line 62: This is optional 🙌 . I feel it inconvenient to access Assets of color & image types a little bit 😄 . I think we can improve it a bit by:

``` Swift
typealias Colors = Assets.Colors
typealias Images = Assets.Images

=> Colors.lightSilver.swiftUIColor
```

nkhanh44: the same with .font 😄 below, I think it could be improved. But it's also fine to be kept as the current one. Not big deal :D.

nkhanh44 (commented):
Great work, just a few comment 🙏 .

.github/workflows/automatic_pull_request_review.yml
```diff
@@ -48,5 +34,8 @@ jobs:
       env:
         CI: true
 
+    - name: Lint
+      run: make lint
+
```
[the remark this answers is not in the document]
Thieurom: @minhnimble Thanks for noticing this!

Actually, if we just run `swiftlint` (warning without fail) the order of `Lint` and `Build` jobs doesn't matter. But since we run it with `--strict` parameter, it's better to put the `Lint` job first so we don't waste time to build if there's linting errors 🙏 

Will update soon 👍

CryptoPrices/Sources/Home/Sources/Home/Localization/Strings.swift
```diff
@@ -1,20 +1,61 @@
...
     }
+    internal enum SeeAll {
+      /// see all ->
+      internal static let text = Strings.tr("Localizable", "home.seeAll.text", fallback: "see all ->")
+    }
+    internal enum Title {
+      /// Localizable.strings
+      ///   CryptoPrices
+      /// 
+      ///   Created by Mike Pham on 06/12/2022.
```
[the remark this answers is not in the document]
Thieurom: This is generated by `swiftgen`. I believe it copies from the original `Localizable` file. Remove this from orignal file and re-generate will remove this too 👍

CryptoPrices/Sources/Home/Sources/Home/WalletStatisticSection.swift
minhnimble: one upvote from me to use `typealias` for cleaner code 👍

CryptoPrices/Sources/Home/Sources/Home/Localization/Strings.swift
Thieurom: @nkhanh44  All this is generated by `swiftgen` by its defaul template. So while this is redundant we don't want to waste effort to edit the files that might be re-written in anytime 🙏

nkhanh44: Yeah, thank you. Swiftgen's template looks a bit messy 😄.

Thieurom:
> Can we add some instructions usage on the new tool - `swiftgen` here? 🙏 This is because when I checkout the source code and add a new localized string, the `Strings` file doesn't get generated automatically with the new contents. 😅 And I suspect the same thing would happen for `Colors` and `Fonts` so please help check it 🙇‍♂️

@minhnimble Actually I added a command `generate` in the `Makefile` in order to quickly re-generate when we add, update resources, but without any information or instructions 🙇

I think I'll make a quick and short instruction to the project's `README` 🙏

CryptoPrices/Sources/Home/Sources/Home/WalletStatisticSection.swift
Thieurom: For the `Colors` and `Images` I intended to make them under the namespace `Assets` since they are actually the types of _assets_ (the resources are even put into files with the `xcassets` extension)
But now I see them quite redundant and lengthy too 😄 

Will update the `swiftgent.yml` to generate without the `Assets` namespace (so no need to add `typealias`)

Thieurom: For the font could you elaborate more 🙏 

As I see now, syntax like `.font(Fonts.Inter.bold)` is quite not too short but is fine (to me) and descriptive. 
This is because it's normal the app contains more than 1 font family (our simple app may contains `Inter` only)
So I think we better explicitly here instead of removing the font name (to become `.font(Fonts.bold)` or `.font(.bold)`)

minhnimble (commented):
Just a final suggestion left, the rest LGTM, really nice work 🚀

nkhanh44 (approved):
🚀

README.md
```diff
@@ -1,12 +1,27 @@
...
+   ```
+   make setup
+   ```
+
+3. Open `CryptoPrices.xcodeproj` with Xcode, hit `⌘R` to run.
+
+4. When adding/removing/updating any resources (colors and images assets, fonts, localized strings), please run the following before re-running the app from Xcode:
+   ```
+   make generate
+   ```
```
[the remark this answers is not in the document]
Thieurom: @minhnimble Yes, I feel that too 🥲

Usually, when we put the resources in the app, we can just add a script in Xcode's build phase so that it can run every time we build (or run) by Xcode. In our case, however, we distribute resources in multiple packages. Of-course we can still do that but it's very strange that to generate code for a module we don't build that module itself but the whole app altogether.

Fortunately, there're some ways 😄:
- Run `swiftgen` every time a folder is changed. This is [documented](https://github.com/SwiftGen/SwiftGen/blob/stable/Documentation/Articles/Watch-a-folder-for-changes.md) by `swiftgen`. This is nice but introduces another dependency ([kicker](https://github.com/alloy/kicker) if we follow their recommendation. It's a Ruby gem, and runs like a daemon server in the background)
- At WWDC 2022, Apple introduced [Swift Package plugins](https://developer.apple.com/videos/play/wwdc2022/110359/) which allow developers to perform actions on Swift packages (which include pre-build action). It's available from Xcode 14 and Swift 5.6. This is exactly what we need and should go for. We can write our own plugin or use one written by someone else [here](https://github.com/SwiftGen/SwiftGenPlugin).

But I think this deserves a dedicated task since we'll have to explore more on Apple documentation and decide to write our own or use a third-party plugin. What do you think 🙏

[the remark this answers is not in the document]
minhnimble: @Thieurom nice investigation 👍 I thought we can just simply add a script in Xcode's build phase but yeah you made valid points there! Let's work on a separate task to handle [Swift Package plugins](https://developer.apple.com/videos/play/wwdc2022/110359/) then. 🚀

minhnimble (approved):

PR #56, opened 2022-12-16
[#10] [UI] As a user, when selecting a coin on the Home screen, I can see the coin’s header information on the Detail screen

### LINKED ISSUE #10: [UI] As a user, when selecting a coin on the Home screen, I can see the coin’s header information on the Detail screen
## Why

- We need to show the coin’s header information on the Detail screen.

## Acceptance Criteria

On the Detail screen, the following section can be seen:

_In light mode:_

Detail Page

_In dark mode:_

Detail Page

* The title will be based on the name from API (handled in integrate story).
* A heart icon will be shown in the top right.

## Design
- [Figma design](https://www.figma.com/file/Iy9eg3ym29pkKGsD5X48O7/Crypto-App-(Community)?node-id=3%3A48&t=PLQA07bsooNZNzKn-0).

nkhanh44:
- Close #10 

## What happened 👀

- Implemented UI header for a Detail screen.
- Added corresponding icons for the back and the like button.

## Insight 📝

- Each feature (Home and MyCoin) has its own state. When the state's changed, AppCoordinator will decide to do something like navigate to the MyCoin screen.
- The implementation of AppCoordinator refers from [this](https://github.com/Thieurom/ic-surveys-ios/pull/44/files#diff-5e09f394864d9cdaf6ff945b44878cddba8810f3faf6fc62289a8d15d8cd7f0b). Thanks to @Thieurom 🙏 .

## Proof Of Work 📹

_In the light mode:_

_In the dark mode:_

_Video:_

https://user-images.githubusercontent.com/25881847/208383962-1b1c012c-98cd-4857-9b1a-1bd9996241d2.mp4


CryptoPrices/Sources/App/AppCoordinator.swift
```diff
@@ -0,0 +1,48 @@
...
+                    return myCoinState
+                }
+                return nil
+            }
+            .flatMap(\.$didSelectBack)
+            .filter { $0 }
+            .sink(receiveValue: { [weak self] _ in
+                self?.state = .home(HomeState())
+            })
+            .store(in: &cancellables)
```
minhnimble, line 46: I am wondering about the scalability of this implementation as for now there is just a simple navigation between 2 screens and we need to handle a lot of navigation code already. Thus, it looks like there is going to be a lot of implementations for this state in this `AppCoordinator` file the more screens we have. 😅

What do you guys think about this [approach](https://quickbirdstudios.com/blog/coordinator-pattern-in-swiftui/) that we involves the coordinator directly into each View's viewModel instead of encapsulating all the states at the App level like the current implementation? This allows better decoupling and make our individual scenes completely independent from their view context. 🙏

CryptoPrices/Sources/App/CryptoPricesApp.swift
```diff
@@ -6,15 +6,33 @@
...
+            HomeView()
+                .environmentObject(homeState)
+        case let .myCoin(myCoinState):
+            MyCoinView()
+                .environmentObject(myCoinState)
+                .transition(
+                    .move(edge: .trailing)
+                    .animation(.linear(duration: 1.0))
+                )
+        }
```
minhnimble, line 30: We are going to have a long list of screens here the more screens we have I believe 😅

CryptoPrices/Sources/Home/Sources/Home/MyCoins/MyCoinSection.swift
```diff
@@ -30,6 +32,13 @@ struct MyCoinSection: View {
                     ForEach(0..<10) {_ in
                         MyCoinItemView()
                             .padding(.trailing, 8.0)
+                            .onTapGesture {
```
minhnimble, line 35: I think we also need to handle the `onTapGesture` to show the Coin Detail screen for the `TrendingCoinsSection` as there are coin item that can be selected too 🙏

nkhanh44: I fixed it at https://github.com/nimblehq/nimble-crypto-ios/pull/56/commits/fad45b30c47233077c24d1567966fd76539f1aa6 🙏

CryptoPrices/Sources/Home/Sources/Home/MyCoins/MyCoinSection.swift
```diff
@@ -30,6 +32,13 @@ struct MyCoinSection: View {
...
+                                DispatchQueue.main.async {
```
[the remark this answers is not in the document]
nkhanh44: Nice catch! I fixed it at https://github.com/nimblehq/nimble-crypto-ios/pull/56/commits/fad45b30c47233077c24d1567966fd76539f1aa6 🙏 .

CryptoPrices/Sources/App/CryptoPricesApp.swift
Thieurom: As I said ☝️, a long list of `switch` here when we have multiple screens/scenes in our `enum Scene` is naturally normal as it's the navigation code 🙏

CryptoPrices/Sources/App/AppCoordinator.swift
nkhanh44: Thanks for the raise @minhnimble. As Thieu mentioned above, with this navigation we could have a couple of approaches but to satisfy the Home and MyCoin shouldn't know each other and they are in the horizontal layer as we discussed [at here](https://nimble-co.slack.com/archives/C04C7E5PWUQ/p1671424415289429). Therefore, we go with this solution which is the most suitable now 🙏 .

minhnimble (approved):
the rest LGTM, really nice work 🚀

CryptoPrices/Sources/App/AppCoordinator.swift
Thieurom: Thank you for your insights @minhnimble 🙏
Totally agree with you!  This is far from a finished solution. And not only us, developers from the community still struggle to find the best solution for this problem, in both UIKit and SwiftUI. I still think we have to address this sooner or later, with a simple, easy-to-implement approach!

Thieurom (approved):
🚀

Issue #54, opened 2022-12-15
[Chore] Add Dependency Injection framework: Factory


minhnimble:
## Why

Currently, we are not using any frameworks or libraries to apply the `Inversion of Control` design pattern. We need to support it for creating ViewModels, Usecases, Repositories, etc. by using [Factory](https://github.com/hmlongco/Factory) framework (the successor of the [Resolver](https://github.com/hmlongco/Resolver) framework).

## Acceptance Criteria

- Add [Factory library](https://github.com/hmlongco/Factory) into the project's SPM.
- Set up some setup classes and foundation injections for later usage.
- Convert existing initializations to use the new DI library.
 
## Who Benefits?

Developers need to access and manage the ViewModels, Usecases, and Repositories instances while developing the app and feature modules.

PR #57, opened 2022-12-16
[#9] [Backend] As a user, I can see Trending coins on the Home screen

### LINKED ISSUE #9: [Backend] As a user, I can see Trending coins on the Home screen
## Why
- The user should be able to see trending coins on the Home screen.

## Acceptance Criteria
- When the user navigates to the Home screen, the following API is called: `GET /coins/markets`.

Parameters:
```
    Coin ids: ids=bitcoin,ethereum,binancecoin,ripple,cardano,solana,polkadot,near,tron,dogecoin
    Currency: vs_currency=usd
    Items per page: per_page=10
    Items order: order=market_cap_desc
    Page: page=1
    Price changes percentage: price_change_percentage=24h
```

- If the API returns success, the user will see the coins items (handled in integrate story #8).
- If the API returns error, the user will see the error state using default alert with title is `Error`, message's content is from API response  and the confirm button's title is `Ok`.

``` JSON
{
  "error": "Could not find coin with the given id"
}
```

## API Documents
- [Coingecko API document](https://www.coingecko.com/en/api/documentation).

minhnimble:
- Close #9

## What happened 👀

- Create the `TrendingCoinsUseCase` for handling API to get trending coin data with the corresponding `TrendingCoinsParameters` and `TrendingCoinsUseCaseProtocol`.
- Add a new coin route - `.trendingCoins` as `"/coins/markets"`.
- Add unit tests for the UseCase and Repository layers.
- Support `coinIDs` parameter as a string array for this trendingCoins API - `GET /coins/markets`.

## Insight 📝

- Currently, when adding more tests to the `CoinRepositorySpec` file, there is a swiftlint violation on both the CI + local and this file needs to disable `swiftlint` manually on the following rules: `// swiftlint:disable function_body_length closure_body_length superfluous_disable_command` at the top level. 👍
- Since the trendingCoins API - `GET /coins/markets` takes the `ids` parameter as a `String` with multiple coin_ids separated by a `comma`, the conversion from a string array to string is intentionally performed at the latest step with this conversion - `coinIDs.joined(separator: ",")` at `Pilot+CoinAPI.swift` file. The goal is to keep using a string array throughout the application and only do the conversion at the latest step, so in other parts of the app, we can easily manipulate the `coinIDs` string array if needed such as sorting, mapping, etc. 🙏
- For the unit testings structure, I updated the existing tests a bit to reflect our iOS team's latest discussion here: https://github.com/nimblehq/compass/discussions/1048

**Notes:** It looks like no evidence of actually calling the trending coins API on the application can be provided at the moment because we need to setup the dependency injections for generating the use cases and repositories in the feature modules such as `Home` module and `MyCoin` module for triggering the API call. The chore task for DI is created [here](https://github.com/nimblehq/nimble-crypto-ios/issues/54) 🙏

## Proof Of Work 📹

All tests passed:

- Repository layer:
Screenshot 2022-12-16 at 17 58 49

- Usecase layer:
Screenshot 2022-12-16 at 17 57 20


Thieurom (approved):
Great work 💯

nkhanh44 (approved):
Great 🚀 🌔

PR #65, opened 2022-12-21
[#54] [Chore] Add Dependency Injection framework: Factory

### LINKED ISSUE #54: [Chore] Add Dependency Injection framework: Factory

minhnimble:
- Close #54

## What happened 👀

- Add [Factory library](https://github.com/hmlongco/Factory) into the project's SPM.
- Set up some setup classes and foundation injections for later usage via the `Container+Injection.swift` file.
- Convert existing initialization to use the new DI library in `CryptoPricesApp` e.g. `HomeViewModel` injection.

## Insight 📝

During the implementation, I encountered a swiftlint issue on the `HomeViewModelSpect.swift` that is discussed [here](https://nimble-co.slack.com/archives/C04C7E5PWUQ/p1671596381472819). In the end, the fix for now is to update the rules to `// swiftlint:disable function_body_length closure_body_length superfluous_disable_command`. We might need to tackle the root cause to make sure the linting on the CI environment should be the same with the local at a later time. 🙏

I also added grouped folders in the main project for better organization:
- `Dependency Injection`folder: contains  `Container+Injection.swift` file.
- `Navigation` folder: contains `AppCoordinator.swift` file and `AppState.swift` file.

Since our setup for the Navigation is on the `CryptoPricesApp`, there is no proper way to setup the Dependencies Injection on each feature modules that there will be plenty of duplicated library imports and DI setup codes, Therefore, I decide to apply DI at the same place with the current Navigation setup code, which is in the main application. 💪

## Proof Of Work 📹

No visible changes, just under the hood optimization and the application built and ran successfully. 🛠🎉

Screenshot 2022-12-22 at 09 53 13


CryptoPrices/Sources/App/Dependencie Injection/Container+Injection.swift
```diff
@@ -0,0 +1,42 @@
...
+import Home
+import MyCoin
+import NetworkExtension
+import UseCaseProtocol
+import NetworkCore
+import Pilot
+import Repositories
+import RepositoryProtocol
+import Styleguide
+import SwiftUI
```
Thieurom, line 18: Just a minor thing, I believe we can omit these 2 imports `Styleguide` and `SwiftUI` here 🙏

Thieurom (approved):
Really nice work 👍

nkhanh44 (approved):
Great work 🚀

CryptoPrices/Sources/App/Dependencie Injection/Container+Injection.swift
minhnimble: you are right, nice catch! Fixed on [5aaaed2](https://github.com/nimblehq/nimble-crypto-ios/pull/65/commits/5aaaed210a7d1f7c1fbb6c1f0ecc2ef054c5dba3)

PR #66, opened 2022-12-21
[#58] [Chore] Set up a code generator or a mocking library for generating stub, fake, mock types for unit testings

### LINKED ISSUE #58: [Chore] Set up a code generator or a mocking library for generating stub, fake, mock types for unit testings
## Why

In order to unit test a type that depends on other types, we inject those dependencies with the mocked implementations.
Currently, in the project, we manually write them by hand. We need to set up a code generator or a mocking library for generating those mocked implementations so that:
- We have the necessary mocked types with clear and consistent API (exposed properties and methods) available.
- Developers just need to focus on writing unit tests.

We can choose between:
- A general-purpose code generator likes [Sourcery](https://github.com/krzysztofzablocki/Sourcery)
- A library specific for mocking likes [Mockingbird](https://github.com/birdrides/mockingbird), [Cuckoo](https://github.com/Brightify/Cuckoo) or others

## Who Benefits?

Developers who implement features that require accompanying unit tests.

Thieurom:
- Close #58

## What happened 👀

- Normally we would install `sourcery` with `homebrew`, then create a `.sourcery.yml`, and put a script build phase in Xcode. However, considering how we structure and architect the app, that approach won't work (or can but with flaws)
  Because the app will import all local packages, generate codes in the app, then the local package _import_ that generated code s, for e.g to write unit tests. This breaks a dependency rule (kind of circular here)
- Instead, we consider local packages are like mini, independent projects, we place generated codes in their own sources, not in a centralized one in the app or in a separate package like `MocksTypes`.
- Create the `BuildTools` package which is responsible for supporting the build process.
- Add the `SourceryPlugin` plugin for generating mocked types (later we should add the `SwiftgenPlugin` for a similar purpose). This plugin is a `prebuildCommand`, which means that when it applies to a package's target it will run _before_ building that target itself:
  - This is the build log of the `Domain` package:
  Screenshot 2022-12-22 at 11 11 43

  - And this is the build log of the whole app:
  Screenshot 2022-12-22 at 11 10 11

  Note the ___Apply build tool plug-in "SourceryPlugin" to target ... in package ...___ part 🙏 

- Apply the plugin to the `Domain` and the `Data` packages. The convention is as follows:
  - Annotate `// sourcery: AutoMockable` on the types we want to generate
  - Generated type name will prefix with `Mock`, for example, `MockCoinRepositoryProtocol`
  - We generate code in a separate file per type, not in a big and messy `AutoMockable.generated.swift` file. The file name will be the same as the type, for example, `MockCoinRepositoryProtocol.swift`
  - Each package will have a dedicated target with the name prefix with `Mocks` containing all the generated files. Example: `DomainMocks` in the `Domain` package.
  - Replace previously written-by-hand mocked types with auto-generated ones.

## Insight 📝

- The hardest part here is generating codes with the correct `import`(s) so that they compile. So in the templates (`stencil` file):
  ```
  {% for import in argument.autoMockableImports %}
  import {{ import }}
  {% endfor %}
  ```
  In the plugin's implementation:
  ```swift
  let sourceryCommand = Command.prebuildCommand(
      displayName: "Generate mocked types for \(target) target",
          executable: sourceryExecutable.path,
          arguments: [
          "--sources",
              target.directory,
              "--templates",
              templatesPath,
              "--args",
   👉         "autoMockableImports=\([target.name] + target.recursiveTargetDependencies.map(\.name))",
              "--output",
              context.package.directory.appending("Sources", "\(context.package.displayName)Mocks"),
              "--verbose"
          ],
          outputFilesDirectory: context.pluginWorkDirectory
  )
  ```
  The generated code must include the `import` of the original type (protocol), hence `target.name`, and it must include the `import`(s) of their dependencies, which are `targetDependency.name`. An example:
  ```swift
  import Entities

  // sourcery: AutoMockable
  public protocol CoinRepositoryProtocol { ... }
  ```
  Code generated for `MockCoinRepositoryProtocol` has to `import` the target `RepositoryProtocol` which consists the definition of `CoinRepositoryProtocol` and has to `import` the dependent `Entities` too.
  ```swift
  import RepositoryProtocol
  import Entities

  public class MockCoinRepositoryProtocol: CoinRepositoryProtocol { ... }
  ```
  This implementation has a big benefit that we don't have to duplicate a `.swiftgen.yml` in every package. We can't have only `.swiftgen.yml` either because all the `--sources`, `--args`, `--output`, ... parameters are different, and manipulating them in code is much easier.

### Some references:
- [WWDC22 - Create Swift Package plugins](https://developer.apple.com/videos/play/wwdc2022/110401/)
- [Writing Sourcery templates](https://github.com/krzysztofzablocki/Sourcery/blob/master/guides/Writing%20templates.md)
- [spm-build-tools](https://github.com/KarimAlweheshy/spm-build-tools)
- [sourcery-plugins](https://github.com/pol-piella/sourcery-plugins)
- [Binary targets in modern Swift packages](https://www.polpiella.dev/binary-targets-in-modern-swift-packages)

## Proof Of Work 📹

Just a short video on how the code is generated on the `Domain` package (note that I open Xcode with the `Domain` package by tapping `Package.swift` in its directory)

https://user-images.githubusercontent.com/17972674/209074160-63dd388a-ead9-4b62-b8a6-38c6326bba81.mp4


CryptoPrices/Sources/Data/Sources/DataMocks/MockCoinAPIProtocol.swift
```diff
@@ -0,0 +1,83 @@
+// Generated using Sourcery 1.9.2 — https://github.com/krzysztofzablocki/Sourcery
+// DO NOT EDIT
+// swiftlint:disable all
+
+import Foundation
+import NetworkCore
+
+public class MockCoinAPIProtocol: CoinAPIProtocol {
```
minhnimble, line 8: I think we should add this file to `.gitignore` as I am wondering if we really need to register this file to Git or not. This is generated code anyway, right? What do you think? 🤔

CryptoPrices/Sources/Domain/Sources/DomainMocks/MockCoinRepositoryProtocol.swift
```diff
@@ -0,0 +1,84 @@
+// Generated using Sourcery 1.9.2 — https://github.com/krzysztofzablocki/Sourcery
+// DO NOT EDIT
+// swiftlint:disable all
+
+import Foundation
+import RepositoryProtocol
+import Entities
+
+public class MockCoinRepositoryProtocol: CoinRepositoryProtocol {
```
minhnimble, line 9: Ditto here regarding `.gitignore` suggestion 🙏

CryptoPrices/Sources/Domain/Sources/DomainMocks/MockMyCoinsUseCaseProtocol.swift
```diff
@@ -0,0 +1,36 @@
+// Generated using Sourcery 1.9.2 — https://github.com/krzysztofzablocki/Sourcery
+// DO NOT EDIT
+// swiftlint:disable all
+
+import Foundation
+import UseCaseProtocol
+import Entities
+
+public class MockMyCoinsUseCaseProtocol: MyCoinsUseCaseProtocol {
```
minhnimble, line 9: Ditto here regarding `.gitignore` suggestion 🙏

CryptoPrices/Sources/Domain/Sources/DomainMocks/MockTrendingCoinsUseCaseProtocol.swift
```diff
@@ -0,0 +1,40 @@
+// Generated using Sourcery 1.9.2 — https://github.com/krzysztofzablocki/Sourcery
+// DO NOT EDIT
+// swiftlint:disable all
+
+import Foundation
+import UseCaseProtocol
+import Entities
+
+public class MockTrendingCoinsUseCaseProtocol: TrendingCoinsUseCaseProtocol {
```
minhnimble, line 9: Ditto here regarding `.gitignore` suggestion 🙏

minhnimble (commented):
This is a very detailed and well-explained PR, excellent work! There are just a few concerns regarding adding generated file to gitignore instead of pushing them to Git and a conflict issue. 🛠

The rest looks great to me 🎉🚀

CryptoPrices/Sources/Data/Sources/DataMocks/MockCoinAPIProtocol.swift
Thieurom: I was actually thinking about this while implementing, here are my thoughts 🙏:
- I think they are the _generated_ codes but they are an integrity part of the package, bundled into one once they're compiled. Of course, these are our codes (in the sense that written by us), packed in local packages, so we can update any time, build time or run time. But what if we treat them like third-party packages? When building we build from source (if open) or import binary one (if closed), I think most of the time (if not all) they don't generate any new files after we import them to our project.
- I'd like to think from another perspective 😄: Imagine we don't have any code generator tool (`Sourcery` or any other), we would add `Swift` files with those mock types written by hand and still add to version control as normal. Now we have `Sourcery` (thank God) we just have _not_ to do that by hand, but everything else is still the same. The codes still are part of the implementation. They are still the public, exposed types that other packages can import and use.
- For the _conflict_ part, that is the reason I generate code in a separate file per type, not in one big file. So for e.g. 2 developers at the same time add 2 new protocols, say `protocol UseCaseA` and `protocol UseCaseB`. In that case, there would be 2 separate files named `MockUseCaseA.swift` and `MockUseCaseB.swift` being generated, not being added in the same file (even in the same line). Even in the case that they both modify the same existing protocol, the chance of conflict is quite low and much easier to resolve.

That said, I just tried to `gitignore` the generated files in the local machine and the project still builds and runs as before. So either way is perfectly fine to me and I'm happy to update. But I'd to get more of your idea 🙏 What do you think?

nkhanh44: IMO, for a more significant project, putting these files into `.gitignore` would be a better practice. But for this project's scope, both approaches may show little difference in effect :D.

nkhanh44 (approved):
So great work and detailed documentation to me 🙌 .

CryptoPrices/Sources/Data/Sources/DataMocks/MockCoinAPIProtocol.swift
minhnimble: @Thieurom to be honest, I think your thought directions are a way far off of my original intention on why we should add it to `.gitignore` here 🙏 Just to clarify a bit more, the topic is mainly on the `Git` storage usage question that should we push the generated files to the repositories or not? To me, the answer is quite clear that it is unnecessary to push generated code to the Git repo because it will be a waste of space and the generated code can be scaled up super quickly. 🙇‍♂️

For our project, at this scale, @nkhanh44 is right that it is not an immediate concern, but I think we should always strive for best practices at the initial phases so that we don't have to spend extra efforts to clean up later on, that would be my take 😁 Anyhow, I think the decision is still belong to the author to make 👍

minhnimble (approved):

CryptoPrices/Sources/Data/Sources/DataMocks/MockCoinAPIProtocol.swift
Thieurom: @minhnimble  @nkhanh44 Thank you guys, I really appreciate your feedback 🙏 

I decided to put those files in `.gitignore`, meaning that initially, those generated files are not parts of the module. They will be after the first build though. However, there was an error:

Screenshot 2022-12-26 at 09 12 12

This happened because when compiling the `Home` package, it needs the `MockMyCoinsUseCaseProtocol` from the `Domain` package. However, the `Domain` package finishes the build process while the `MockMyCoinsUseCaseProtocol.swift` is generated or saved on the file system. It's kind of _not waiting_ for the files to be done generating. I realized that I had a mistake:
```swift
"--output",
context.package.directory.appending("Sources", "\(context.package.displayName)Mocks"),
```
I put the generated files in a _custom_ directory that is outside of the target that is currently being built (another target name "DomainMocks"). Those files wouldn't be the sources of the target, the compiler wouldn't compile those files. So the error occurred.

Actually, besides the `Sources` directory, each Swift package or target has another directory that the compiler will search while building, that is the `pluginWorkDirectory`. Here is a little more detail on [Package Manager Extensible Build Tools proposal](https://github.com/apple/swift-evolution/blob/main/proposals/0303-swiftpm-extensible-build-tools.md): 
>Every plugin invocation is passed the path of a directory in TargetBuildContext.pluginWorkDirectory. A plugin would usually create a separate subdirectory of this directory for each prebuild command it creates, and the command would be configured to write its output files into that directory. The plugin may also create other subdirectories for cache files and other file system content that either it or the command needs. If these additional files need to be available to the command, the plugin would construct the command line to include their paths, for the command to use.
Examples of plugins that need to use prebuild commands include SwiftGen and other tools that need to see all the input files, and whose set of output files is determined by the contents (not just the paths) of the input files. Such a plugin usually generates just one command and configures it with an output directory into which all generated sources will be written.

Here is an example of the `pluginWorkDirectory` path while building on my machine:

Screenshot_2022-12-26_at_09_52_18

So I updated with the following:
```swift
let generatedDirectory = context.pluginWorkDirectory.appending("SourceryGenerated")
//
"--output",
generatedDirectory,
//
```
Actually, all the references in the PR's description follow this config/approach but I was too hurry to check them carefully 😓
So now it:
- Fix the build issue
- The generated files are not visible to the Xcode's navigator which is good cause we rarely bother with the content or what's going on in them. We can still jump to the definition if we want to. But we can't accidentally modify them (at least in Xcode) which is another good!
- Putting those in `.gitignore` is not necessary anymore since they belong to the sandbox directory.

PR #67, opened 2022-12-22
[#13] [UI] As a user, I can see the coin’s current price on the Detail screen

### LINKED ISSUE #13: [UI] As a user, I can see the coin’s current price on the Detail screen
## Why

- The user should be able to see coin’s current price on the Detail screen.

## Acceptance Criteria

- On the Detail screen, the following section can be seen:

_In light mode:_

Detail Page

_In dark mode:_

Detail Page

- The image will be based on the `image.large`.
- The current price will be based on the `current_price.usd` and the fractional part will be rounded to `2` numbers.
- The price change percentage will be based on the `price_change_percentage_24h`:

  - `price_change_percentage_24h` text is set to `#F15950` with the down indicator and background is set to `#F15950` with alpha `0.3` when it’s a negative number.

  - `price_change_percentage_24h` text is set to `#10DC78` with the up indicator and background is set to `#DBFBEC` when it’s a positive number.

Screen Shot 2022-11-25 at 11 25 29

Screen Shot 2022-11-25 at 11 26 07

_Note:_ The fractional part of `price_change_percentage_24h` rounded to 2 numbers: `6,2122` to `6,21`.

## Design
- [Figma design](https://www.figma.com/file/Iy9eg3ym29pkKGsD5X48O7/Crypto-App-(Community)?node-id=3%3A48&t=PLQA07bsooNZNzKn-0).

minhnimble:
- Close #13

## What happened 👀

- Implement the current price section for the `MyCoin` screen.
- Add a base scroll view for this section to support scrolling with more content in the future. 

## Insight 📝

- Currently, some of the current set fonts and sizes are incorrect compared to the [designs](https://www.figma.com/file/Iy9eg3ym29pkKGsD5X48O7/Crypto-App-(Community)?node-id=3%3A107&t=RTdnDWc1ccmZrfsh-0) that the weight should be `semiBold` and `medium` but we are using `bold` most of the time. Here is the `Inter` font's weight for more reference: https://fonts.google.com/specimen/Inter -> So a refactor for updating to consistent fonts and sizes usage compared with the design is applied for the new screen and the existing screens also 🚀

- Regarding the new color `#DBFBEC`, when trying to get a name from this [source](https://www.color-name.com/hex/d6f5f3), it provides the same name -`water` as with the color - `#D6F5F3`. As a result, I am using the name provided by this [source](https://chir.ag/projects/name-that-color/#DBFBE) instead because it has better naming differentiation between small changes in the color codes. 🙏

- Update the lint to clean up `superfluous_disable_command` since my local Swiftlint is updated to the latest version `0.50.3` and it is working properly now 🚀

## Proof Of Work 📹

- Dark mode:

- Light mode:

### DIFF
--- a/CryptoPrices/Sources/MyCoin/Sources/MyCoin/CurrentPriceSectionView/CurrentPriceSection.swift
+++ b/CryptoPrices/Sources/MyCoin/Sources/MyCoin/CurrentPriceSectionView/CurrentPriceSection.swift
@@ -1,0 +1,47 @@
+//
+//  CurrentPriceSection.swift
+//  MyCoin
+//
+//  Created by Minh Pham on 22/12/2022.
+//
+
+import Styleguide
+import SwiftUI
+
+struct CurrentPriceSection: View {
+
+    var body: some View {
+        VStack(spacing: 8.0) {
+            // TODO: - Remove dummy
+            Images.icEth.swiftUIImage
+                .frame(width: 60.0, height: 60.0)
+                .clipShape(Circle())
+
+            VStack(spacing: 4.0) {
+                // TODO: Remove hard-coded data
+                Text(7_273_291, format: .dollarCurrency)
+                    .foregroundColor(.white)
+                    .font(Fonts.Inter.bold.textStyle(.title2))
+
+                Button(action: {}, label: {
+                    HStack {
+                        Images.icArrowUpGreen.swiftUIImage
+                        Text(2.41, format: .percentage) // TODO: Remove hard-coded data
+                    }
+                })
+            }
+        }
+        .padding(.vertical, 8.0)
+    }
+}
+
+#if DEBUG
+struct CurrentPriceSection_Previews: PreviewProvider {
+
+    static var previews: some View {
+        Preview {
+            CurrentPriceSection()
+        }
+    }
+}
+#endif
--- a/CryptoPrices/Sources/MyCoin/Sources/MyCoin/CurrentPriceSectionView/CurrentPriceSection.swift
+++ b/CryptoPrices/Sources/MyCoin/Sources/MyCoin/CurrentPriceSectionView/CurrentPriceSection.swift
@@ -14,24 +14,35 @@ struct CurrentPriceSection: View {
         VStack(spacing: 8.0) {
             // TODO: - Remove dummy
             Images.icEth.swiftUIImage
+                .resizable()
+                .aspectRatio(contentMode: .fit)
                 .frame(width: 60.0, height: 60.0)
                 .clipShape(Circle())
 
             VStack(spacing: 4.0) {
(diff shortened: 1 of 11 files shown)


CryptoPrices/Sources/MyCoin/Sources/MyCoin/CurrentPriceSectionView/CurrentPriceSection.swift
```diff
@@ -0,0 +1,58 @@
...
+                    .font(Fonts.Inter.semiBold.textStyle(.title))
+                    .frame(height: 34.0)
+
+                Button(action: {}, label: {
+                    HStack {
+                        Images.icArrowUpGreen.swiftUIImage
+                        // TODO: Remove hard-coded data
+                        Text(2.41, format: .percentage)
+                            .font(Fonts.Inter.medium.textStyle(.callout))
+
```
nkhanh44, line 35: I think we can remove this empty line. 🙏

CryptoPrices/Sources/MyCoin/Sources/MyCoin/CurrentPriceSectionView/CurrentPriceSection.swift
```diff
@@ -0,0 +1,58 @@
...
+            // TODO: - Remove dummy
+            Images.icEth.swiftUIImage
+                .resizable()
+                .aspectRatio(contentMode: .fit)
+                .frame(width: 60.0, height: 60.0)
+                .clipShape(Circle())
+
+            VStack(spacing: 4.0) {
+                // TODO: Remove hard-coded data
+                Text(7_273_291, format: .dollarCurrency)
```
nkhanh44, line 24: It's optional. If we have a decimal number (for exp: `7_273.291`) like `2.41` below, it would be nice :D.

nkhanh44 (approved):
Great PR 💯

Thieurom (approved):
🎉 🚀

CryptoPrices/Sources/MyCoin/Sources/MyCoin/CurrentPriceSectionView/CurrentPriceSection.swift
minhnimble: Oh, it is intentional to place a line break here just for better readability, so I would prefer to keep it 😁

CryptoPrices/Sources/MyCoin/Sources/MyCoin/CurrentPriceSectionView/CurrentPriceSection.swift
minhnimble: I think it would not be a big deal, as we will need to update to real data anyway. Let's keep it as-is 🙏

PR #74, opened 2022-12-27
[#71] [Chore] Improve navigation with proper app state management

### LINKED ISSUE #71: [Chore] Improve navigation with proper app state management
### Issue
Currently, from the __Home__ screen, if the users select a coin, it will:
- Display the __MyCoin__ screen. But it does so by removing the __Home__ from the view hierarchy, not by pushing the __MyCoin__ screen on the navigation stacks (on top of the __Home__ screen)
- When on the __MyCoin__ screen, if the users tap the __back__ button, it will destroy the __MyCoin__ screen and _recreate_ the __Home__ screen from scratch. It would make API calls again, and the users lose the previous scroll positions on the __Home__ screen.
- We wrap __Home__ and __MyCoin__ screens in their own`NavigationView`s, not in the same one, and use a custom transition. As the result the animation is not quite matched with the built-in one, it also loses the _back_ animation and swipe-to-back gesture.

  https://user-images.githubusercontent.com/17972674/209495140-25291f06-f675-4020-a352-990ef5ad6811.mp4

### Proposed approach
- We can't model the `AppState` which currently consists of `HomeState` and `MyCoinState` with the `enum`. Because enum cases are exclusive to each other. When we change `state` from `HomeState` to `MyCoinState`, the __Home__ screen would be destroyed.
- We should model the `AppState` in a way that the `HomeState` is always there, and the `MyCoinState` is the `Optional` type indicating that when we can unwrap its value, the __MyCoin__ screen will be presented.
- Put both the __HomeView__ and __MyCoinView__ in the same `NavigationView` with the __HomeView__ being the root one to get the benefits of smooth built-in transitioning animations.

## Who Benefits?

Users, developers.

Thieurom:
- Close #71

## What happened 👀

- Model `AppState` as a struct instead of an enum.
- Create `AppView` which is responsible to create the whole view hierarchy.
- Remove `NavigationView` from the `HomeView` and `MyCoinView`.
  Instead, create only one in the `AppView` with the `HomeView` as the root view, and programmatically trigger navigation to the `MyCoinView`.
- Handle tapping on the custom back button and swiping left-to-right on the `MyCoinView`.
- Refactor `AppCoordinator` a little bit with the `assign` operator (so no need for a set of `AnyCancellable` anymore)

## Insight 📝

SwiftUI in iOS15 is still limited in terms of navigation.

## Proof Of Work 📹

| Before | After |
|---|---|
|<video src="https://user-images.githubusercontent.com/17972674/209607239-e67edd55-2653-478b-968b-0d37a88a48d1.mp4">|<video src="https://user-images.githubusercontent.com/17972674/209607242-8b7eb750-bb27-431d-89e1-dc73c12cd402.mp4">|


nkhanh44 (approved):
Really nice PR. It works correctly now :D. 🚀

CryptoPrices/Sources/Home/Sources/Home/TrendingCoinsSectionView/TrendingCoinSection.swift
```diff
@@ -29,13 +29,12 @@ struct TrendingCoinSection: View {
             // TODO: - Remove dummy
             ScrollView(.vertical, showsIndicators: false) {
                 VStack {
-                    ForEach(0..<2) {_ in
+                    ForEach(0..<2) { _ in
```
minhnimble, line 32: nice spot 👍

CryptoPrices/Sources/Home/Sources/Home/TrendingCoinsSectionView/TrendingCoinSection.swift
```diff
@@ -29,13 +29,12 @@ struct TrendingCoinSection: View {
...
                         TrendingCoinItemView()
                             .padding(.vertical, 8.0)
                             .onTapGesture {
-                                withAnimation {
-                                    homeState.didSelectCoin = true
-                                }
+                                // TODO: Remove hard-coded data
+                                homeState.didSelectCoin = "bitcoin"
```
minhnimble, line 37: Oh this is intentional that we remove `withAnimation` and we don't need it anymore, right? 

When presenting the view, I can see that there is still a proper transition, even for the going back action, nice work 🎉

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
```diff
@@ -12,13 +12,20 @@ public struct MyCoinView: View {
...
+            .navigationTitle("Ethereum") // TODO: Remove dummy
+            .navigationBarTitleDisplayMode(.inline)
+            .gesture(
+                DragGesture()
+                    .onEnded { value in
+                        // Handle swipe left-to-right
+                        if value.startLocation.x < 20.0 && value.translation.width > 50.0 {
+                            myCoinState.didSelectBack = true
+                        }
+                    }
```
minhnimble, line 27: Hmm I don't think there is a requirement for swiping back to go back to the previous screen but yeah it is a nice to have thing 👍 

Just a thought, maybe we can consider creating a view extension for this swipe-back gesture so we can apply it for multiple screens when the project scales up, as of now it is still good to just apply it here locally for `MyCoinView` 🙏

minhnimble (approved):
Just a small suggestion and a double-checking from me, the implementation itself looks great already, nice work 🚀

CryptoPrices/Sources/Home/Sources/Home/TrendingCoinsSectionView/TrendingCoinSection.swift
Thieurom: Yeah, the `withAnimation` is no longer necessary because we don't manually add animation when transitioning from `Home` to `MyCoin` but let the `NavigationView` do it for us.

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
Thieurom: Thank you 👍

For now, I think I'd apply locally for `MyCoinView`. Later on we can make it as an extension as your suggestion 🙏
Note that in that case, it would be placed in a separate package, i.e. `Styleguide` in our case.

PR #75, opened 2022-12-27
[#39] [Backend] As a user, I can see the chart on the Detail screen

### LINKED ISSUE #39: [Backend] As a user, I can see the chart on the Detail screen
## Why
- The user should be able to see the chart on the Detail screen.

## Acceptance Criteria
- When the user navigates to the Detail screen, the following API is called for 1D price: GET /coins/{id}/market_chart.

Parameters:
> id: pass the coin id
vs_currency: usd
days: 1

```JSON
{
  "prices": [
    [
      1669284289704,
      16595.246079387816
    ],
    [
      1669284585059,
      16602.076684659474
    ],
    [
      1669284921814,
      16606.11839794301
    ],
    [
      1669285188016,
      16606.20441243831
    ]
}
```

* If the API returns success, the user will see the chart (handled in integrate story).
* If the API returns an error, the user will see the error state using the default alert with the title Error, the message's content is from the API response, and the confirm button's title is Ok.

``` JSON
{
  "error": "Could not find a coin with the given id"
}
```

## API Documents
- [Coingecko API document](https://www.coingecko.com/en/api/documentation).

minhnimble:
- Close #39

## What happened 👀

- Create the new Repository functions and `GetChartPricesUseCase` along with all the necessary protocols.
- Create new models for parsing data with cross-application usage: `APIPrices`, `DataPoint`, `ChartDataPoint`, `TimeFilters`, etc.
- Create a supporting enum - `TimeFilter` that can help provides different days range for API usage:
  - 1D = 1 day
  - 1W = 7 days
  - 1M = 30 days
  - 1Y = 365 days
  - 5Y = 1285 days (365*5)
- Add unit tests for Repository and UseCase layers.

## Insight 📝

- No major highlights, just reusing the established architecture and adding new `UseCase` and `Repository` functions for getting chart data.

- There is a small notice that the `TimeFilters` enum is decided to be placed inside the `Domain` package instead of the `Data` package because of our current dependencies setup. The `Domain` package is currently imported into the `Data` package so if we implement the enum in `Data`, it will create a recursive importing cycle - `Domain -> Data -> Domain`. This will lead to a build failed issue prompted by Xcode. ⚠️
Screenshot 2022-12-27 at 16 18 19

## Proof Of Work 📹

All tests passed:

- Repository layer:
Screenshot 2022-12-27 at 16 00 59

- Usecase layer:
Screenshot 2022-12-27 at 16 03 53


CryptoPrices/Sources/Domain/Sources/Entities/TimeFilter.swift
```diff
@@ -0,0 +1,24 @@
+//
+//  TimeFilter.swift
+//  Domain
+//
+//  Created by Minh Pham on 27/12/2022.
+//
+
+public enum TimeFilter {
+    case oneDay
```
nkhanh44, line 9:
```suggestion
public enum TimeFilter {

    case oneDay
```

CryptoPrices/Sources/Data/Sources/NetworkExtension/CoinRoute.swift
```diff
@@ -24,6 +25,7 @@ extension CoinRoute: Route {
     public var path: String {
         switch self {
         case .myCoins, .trendingCoins: return "/coins/markets"
+        case .chart(let parameters): return "/coins/\(parameters.id)/market_chart"
```
nkhanh44, line 28: For consistency, let's use the following suggestion:
```suggestion
        case let .chart(parameters): return "/coins/\(parameters.id)/market_chart"
```

nkhanh44 (approved):
Nice work! just some minor changes. So it's ready to 🚀 .

CryptoPrices/Sources/Domain/Sources/Entities/TimeFilter.swift
minhnimble: Fixed in [4338b7e](https://github.com/nimblehq/nimble-crypto-ios/pull/75/commits/4338b7e9fe2efa557c9c644bfc0289722e2561ce) 🚀

CryptoPrices/Sources/Data/Sources/NetworkExtension/CoinRoute.swift
minhnimble: Fixed in [4338b7e](https://github.com/nimblehq/nimble-crypto-ios/pull/75/commits/4338b7e9fe2efa557c9c644bfc0289722e2561ce) 🚀

CryptoPrices/Sources/Data/Sources/NetworkExtension/Parameters/GetChartPricesParameters.swift
```diff
@@ -0,0 +1,13 @@
...
+//  Data
+//
+//  Created by Minh Pham on 27/12/2022.
+//
+
+public struct GetChartPricesParameters: Parameterable {
+
+    public let id: String
+    public let vsCurrency = "usd"
+    public let days: String
```
Thieurom, line 12: The `days`'s type should be `Int` 🙏

CryptoPrices/Sources/Data/Sources/NetworkExtension/Pilot+CoinAPI.swift
```diff
@@ -5,6 +5,7 @@
 //  Created by Doan Thieu on 01/12/2022.
 //
 
+import Entities
```
Thieurom, line 8: We don't have to import `Entities` anymore 🙏

CryptoPrices/Sources/Data/Sources/Repositories/CoinRepository.swift
```diff
@@ -8,6 +8,7 @@
 import Entities
 import NetworkCore
 import RepositoryProtocol
+import NetworkExtension
```
Thieurom, line 11: The `Repository` can work with any type that conform to `CoinAPIProtocol` (hence the need of `import NetworkCore`) It doesn't depend on a specific implementation. So we should omit the `import NetworkExtension` here.

CryptoPrices/Sources/Data/Package.swift
```diff
@@ -35,13 +35,16 @@ let package = Package(
...
+            dependencies: [
+                .product(name: "Entities", package: "Domain")
+            ],
             plugins: [.plugin(name: "SourceryPlugin", package: "BuildTools")]
         ),
         .target(
             name: "NetworkExtension",
             dependencies: [
                 "NetworkCore",
+                .product(name: "Entities", package: "Domain"),
```
Thieurom, line 47: This can omit too 🙏

CryptoPrices/Sources/Data/Sources/NetworkCore/APIModels/APIPrices.swift
```diff
@@ -0,0 +1,13 @@
...
+//
+//  Created by Minh Pham on 27/12/2022.
+//
+
+import Foundation
+
+public struct APIPrices: Decodable, Equatable {
+
+    public let prices: [[Double]]
+}
```
Thieurom, line 13: For the price, I'd prefer to use the `Decimal` type than `Double`. Although our app does not involve any calculations so we don't have to worry about like miscalculation, rounding errors, ... but it's a good convention I think. And to make it consistent too since other types use `Decimal` already like `Coin` and `APICoin`.

To make the consumer of the network layer's life a little easier,  we can let the `CoinAPIProtocol` return a more usable types:
```swift
public protocol CoinAPIProtocol {

    func getChartPrices(coinID: String, numberOfDays: Int) async throws -> [APIPrice]
}
```
The remote API returns data in a format `prices: [[String]]`, so we can add an intermediate data type:
```swift
// Remote API returns
public struct APIPriceData: Decodable, Equatable {

    public let prices: [[Decimal]]
}
```

This is what the network layer returns:
```swift
// Our network layer returns (to Repository)
public struct APIPrice: Equatable {

    public let price: Decimal
    public let timestamp: Date

    public init(price: Decimal, timestamp: Date) {
        self.price = price
        self.timestamp = timestamp
    }
}
```

The implementation of the network layer just do some small extra work to convert to correct type:
```swift
extension Pilot: CoinAPIProtocol where R == CoinRoute {

    public func getChartPrices(coinID: String, numberOfDays: Int) async throws -> [APIPrice] {
        let priceData = try await request(
            .chart(GetChartPricesParameters(id: coinID, days: numberOfDays)),
            target: APIPriceData.self,  👈  Raw, original from remote sources
            decoder: .apiDecoder
        )

        return priceData.prices.map { prices in
            APIPrice(
                price: prices.first ?? 0.0,
                timestamp: prices.last.map { Date(timeIntervalSince1970: ($0 as NSDecimalNumber).doubleValue) } ?? Date()
            )
        }
    }
}
```
We can safely remove the `DataPoint` from the network layer 🎉

CryptoPrices/Sources/Data/Sources/Repositories/CoinRepository.swift
```diff
@@ -25,6 +26,14 @@ public class CoinRepository: CoinRepositoryProtocol {
...
     }
 
+    public func getChartPrices(coinID: String, filter: TimeFilter) async throws -> [ChartDataPoint] {
+        let chartData = try await coinAPI.getChartPrices(coinID: coinID, filter: filter)
+        let dataPoints = chartData.prices.map {
+            DataPoint(timestamp: $0.first ?? 0.0, price: $0.last ?? 0.0)
+        }
+        return dataPoints
+    }
+
```
Thieurom, line 36: This is where we can transform from higher abstraction to lower one:
```swift
public func getChartPrices(coinID: String, filter: TimeFilter) async throws -> [ChartDataPoint] {
  return try await coinAPI.getChartPrices(coinID: coinID, numberOfDays: filter.daysCount).map { $0 as ChartDataPoint }
}

extension APIPrice: ChartDataPoint {}
```

It lies on the `numberOfDays: filter.daysCount` (`TimeFilter` to `Int`) 🎉

Thieurom (commented):
This is already a great work 👍 

I had some concerns and suggestions, please consider them and feel free to discuss further 🙏 

As a side note, I think modularization force us to think carefully how we modelize the abstractions, what we expose, what we should hide, what we depend, ... The compiler will help us here like error out the missing type (need `import`) or `cyclic dependency` as you faced.
Normally when we place all the sources code in one place, it's so convinient to just use any type from any location without realizing it violates dependency rule, or separation of concerns, ...

CryptoPrices/Sources/Data/Sources/NetworkExtension/Parameters/GetChartPricesParameters.swift
minhnimble: I am not sure why you raised this but the `day`'s type from the Coingecko [document](https://www.coingecko.com/en/api/documentation) is `String`, not `Int` 😅

Screenshot 2022-12-28 at 15 14 46

CryptoPrices/Sources/Data/Sources/NetworkCore/Models/DataPoint.swift
```diff
@@ -0,0 +1,19 @@
...
+public struct DataPoint {
+
+    public let timestamp: TimeInterval
+    public let price: Double
+
+    public init(timestamp: TimeInterval, price: Double) {
+        self.timestamp = timestamp
+        self.price = price
+    }
+}
```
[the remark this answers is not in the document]
minhnimble: @Thieurom this model is not a direct map of the API's response (the response doesn't contain any fields that have `timestamp` and  `price`), so that is why I intentionally remove the word `API` from the model 😅 And thus, I don't think this class can be replaced with `APIPrice` from your suggestion, due to another reason that is `Price` is not the right naming for a model that contains timestamp and price, logically it is a chart data point 🙏

CryptoPrices/Sources/Data/Sources/NetworkCore/CoinAPIProtocol.swift
```diff
@@ -5,11 +5,14 @@
...
+import Entities
+
 // sourcery: AutoMockable
 public protocol CoinAPIProtocol {
 
     // TODO: Update when implement
     func myCoins() async throws -> [APICoin]
     func trendingCoins(coinIDs: [String]) async throws -> [APICoin]
     func coinDetail() async throws -> APICoin
+    func getChartPrices(coinID: String, filter: TimeFilter) async throws -> APIPrices
```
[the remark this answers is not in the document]
minhnimble: > I think we can still make the network layer work without depending on the Domain layer here.

-> To be honest with you, please correct me if I am wrong but I am quite confused with this claim because we are already making the `Data` layer depend on the `Domain` layer, no? Practically, it is impossible to completely separate them and I actually feel it is natural to let the `Data` layer depend on the `Domain` layer. 🙏
For example, we have `RepositoryProtocol`  inside the `Domain` layer, and Repositories in the `Data` layer need to conform to the protocols, right?
Screenshot 2022-12-28 at 15 31 06

> We just provide directly a `numberOfDays`, we don't have to use higher abstraction like a custom enum here. The parameter needs update too.

-> In the beginning, I had the same thought as you to just provide `numberOfDays` as `Int` too 👍 Then, when thinking in the long run, the main purpose of introducing `enum` here is for reusing it in the future in the logic layer - `ViewModel` or even the `View` layer (since we are using SwiftUI). Later on, when we need to switch between different filters to call the correct API and draw different charts, now we already have the enum representation to work with. And the day's calculation introduced in the enum is just a technique to group related code for a more descriptive way of usage. 🙏

> The function name, parameter types, and return types between different layers are not necessarily the same and actually have nothing to do with others.

-> It is my intention to use parameter types as `enum` for a more descriptive type of writing code instead of just providing plain numbers and we have to add extra comments/descriptions for it at a later time in the Network layer here for protocols. Please correct me but I still don't think there is a valid concern to stop using `enum` at this point.   🙇‍♂️ Why do we have to stick with the `low-level`, `raw data` type usage for protocols? 🤔

CryptoPrices/Sources/Data/Sources/Repositories/CoinRepository.swift
minhnimble: nice catch! I was testing different approaches on how to manage the shared enum with the current modular architecture but I forgot to clean up this one, will remove 👍

CryptoPrices/Sources/Data/Sources/NetworkCore/APIModels/APIPrices.swift
minhnimble: One quick question, at the `CoinAPIProtocol` layer, shouldn't we return the exact response object and the mapping can be done via the `Repository` layer? I raised this because `APIPrice` is not an identical mapping to the response payload and thus it should not be treated as an `API` response IMO 🙏

On top of that, regarding the suggestion of the `APIPrice` class, I have the same concern as [this comment](https://github.com/nimblehq/nimble-crypto-ios/pull/75#discussion_r1058146716) 😅

> For the price, I'd prefer to use the Decimal type than Double. Although our app does not involve any calculations so we don't have to worry about miscalculations, or rounding errors, ... but it's a good convention I think. And to make it consistent too since other types use Decimal already like Coin and APICoin.

I have no issue with this and your reason makes sense to me, I will update 👍

CryptoPrices/Sources/Data/Sources/Repositories/CoinRepository.swift
minhnimble: As per the explanation from my [comment](https://github.com/nimblehq/nimble-crypto-ios/pull/75#discussion_r1058163342), I don't see any major concern to stop using enum here. Furthermore, looking at your suggestion, I think we should try to avoid complex operations as per our [convention](https://nimblehq.co/compass/development/ios/build-time-optimization/) here, and thus, we should create a `let` constant per each operation instead. 🙏

CryptoPrices/Sources/Data/Sources/NetworkCore/APIModels/APIPrices.swift
minhnimble: Update to use `Decimal` in [2378259](https://github.com/nimblehq/nimble-crypto-ios/pull/75/commits/2378259dd05125751a71af9148163c515fd36042) 🚀

CryptoPrices/Sources/Data/Sources/NetworkExtension/Parameters/GetChartPricesParameters.swift
Thieurom: Oh, I missed this part 🙏

But if you really think about it, technically `String` can be used to represent _number of days_, but semantically `Int` would be the better choice. I don't know the intention of the CoinGecko here. The other API like `coins/market` has the parameters with the type of `Int`:

Screenshot 2022-12-29 at 09 15 43

They could've used `String` for `per_page` and `page` here as they did for `days` but they didn't. My best guess is this has something to do with the types they use in their database schema, which may have some history 😄 

Put that aside, speaking of implementation, either `Int` or `String` works cause we use string interpolation when encoding the parameters (i.e. `"\(days)"`. But it's fair and practically correct to use `String` here to match and keep it in sync with the CoinGecko API 🙏

CryptoPrices/Sources/Data/Sources/NetworkCore/Models/DataPoint.swift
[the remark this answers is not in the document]
Thieurom: Yeah, I got your point here 🙏

But as you mentioned,
>the `DataPoint` is not a direct map of the API's response

and 
> shouldn't we return the exact response object and the mapping should be done via the Repository instead

why bother putting it in the network layer? Why don't put this `DataPoint` directly in the `Repositories` module? I tried and worked fine.

Screenshot 2022-12-29 at 11 25 31

That is because `DataPoint` is a kind of intermediate type for transforming data purposes, used specifically on the implementation detail of the `Repositories`. At the end of the day, what it returns is still the abstract type (`CharDataPoint` protocol) by casting. So the dependee (`GetChartPricesUseCase`) doesn't know that. You can omit the `public` marks on the definition of `DataPoint` too to make it internally to `Repositories`.

CryptoPrices/Sources/Data/Sources/NetworkExtension/Parameters/GetChartPricesParameters.swift
Thieurom: [Updated] Now I realized the `days` parameter may accept a `"max"` value, which may cap the maximum ranges that they can return.  Finally `String` makes sense to me 😄

minhnimble: @Thieurom yeah, you are totally right that this `day`'s type should be `Int` and it would definitely make more sense than `String` here. Unfortunately, CoinGecko didn't do it for this API 🙈

So I will have to blindly stick with their API interface, what a pain 🤷‍♂️

CryptoPrices/Sources/Data/Sources/NetworkCore/APIModels/APIPrices.swift
Thieurom: Yeah, totally agree with you here! The transformation should be made in the `Repository` layer 🙏

CryptoPrices/Sources/Data/Sources/NetworkCore/CoinAPIProtocol.swift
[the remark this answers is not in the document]
Thieurom: @minhnimble Thank you for your detailed feedback 🙏  I'd like to continue sharing my concerns here:
- About using the `TimeFilter` enum: I think you missed my point. I didn't say to not use the enum, but rather the opposite:
  > So it's great that on the Domain layer we can provide the TimeFilter type that abstracts away the number of days with a set of predefined ones.

  You're encouraged to use it in the higher layers like `Repository`, `UseCase`, or even better in the `ViewModel` and `View` which are closer to the presentation layer. What I meant was you shouldn't use that enum in the network layer. Here're 2 obvious reasons:
  - The `CoinAPIProtocol` currently can do `func getChartPrices(coinID: String, filter: TimeFilter) async throws -> APIPrices`. Meaning that it _only can make the request with the days that are one of the values 1, 7, 30, 365, and 1285_. What if I want to get the history price for 3, 57, 2589 days, or any other given day? I can't, right? By forcing it to accept only a small set of values, you limit the network layer's functionality and capability.
  - Using the enum here not only limits the network's capability, but it also becomes too coupled with the app business. Since the `MyCoinView` display the chart with a predefined set of days, you modelize the network in a way to support just that.
 
    What if the `MyCoinView` can also display data for like ten years? You have to add a new case for `TimeFilter` and thus have to update the network 🤔?

    Another example is that suppose our app allows users to register an account, with 2 tiers: free and premium. A free account can only see the historical prices for what is defined in `TimeFilter`. A premium account has an additional option to choose whatever day ranges they want to see (for e.g with a Slider on the UI with a value between 1 and 10_000). In that case, updating `ViewModel`, and `UseCase` is logical and necessary. But modify the network layer to make it work? It's supposed to be updated only on some rare cases like CoinGecko API breaks something, or we update from `APIv1` to `APIv2`, etc,... But it _never updates when the app business changes, or when the UI/UX changes_. You have to agree with me it's strange if it would be the case right 🙏

   As a side note, we can use the enum in the network layer when the CoinGecko API accepts those, and a very good candidate is the `vs_currency` parameter. They accept values like `usd`, `eur`, `jpy`, etc, ... They must keep a valid set on the database schema so the caller can't just make a request with an arbitrary string like `lorem ipsum`.

- About the relationship between layers, the `Data` _does_ depend on the `Domain` in some senses, but it's more complicated than that. In SPM terms, the _package_ is just kind of a container (it may be oversimplified but I think you got the point). What is matter here is the modules that the package exposed. So the `Data` _package_ has 4 _modules_:
  - `NetworkCore`
  - `NetworkExtension`
  - `Repositories`
  - `DataTestHelpers`
  
  while the `Domain` _package_ has following _modules_:
  - `Entities`
  - `RepositoryProtocol`
  - `UseCaseProtocol`
  - `UseCases`
  - `DomainTestHelpers`

  As you indicated, the `Repositories` module (concrete type) of the `Data` depends on the `RepositoryProtocol` (interface) of the `Domain`. What I meant when I say `can still make the network layer work without depending on the Domain layer` is that you don't have to make the `NetworkExtension` depend on the `Entities` (to use the `TimeFilter`). Of course, you can but with the ☝️ above point, you shouldn't and don't have to.

Last but not least, I'd like to take this chance to present a clear _picture_ of how we structure and architect the app so far. Why? Because a picture is worth a thousand words 🙏 :

CryptoPricesArchitecture

(Don't judge my drawing skills. Don't even ask me about UML stuff either 😄 I'm very bad at those kinds of things)
We modulize the app by distributing the sources into different layers. Each layer contains one or more packages. Each package contains one or more modules that can be `import`ed by others.
Some worth-to-mention points here (please take your time to check this later, and feel free to raise you ideas):
- Layers are organized vertically. A module of a lower layer depends on one or more modules of the higher layer, but not vice versa.
- Modules from packages on the same layer don't know each other. For e.g. the `Home` vs the `MyCoin` in the __Features__. Or the `BuildTools` vs the `Styleguide` in the __Utilities__.
- A concrete type of module on a lower layer depends on the abstraction (interface) of a higher layer. For e.g. the `Repositories` (an implementation of the repository) depends on the `NetworkCore`. So when we change the implementation of the network (`NetworkExtension`),  say we switch from `Pilot` to `Moya` and `Alamofire`, that `Repositories` won't change and won't be re-built when the app builds. And you definitely know that this has the benefit for unit testing too 🙏 
- The app is the one that uses mostly concrete types. This is obvious because the implementation is what makes the app does what it advertises (business logic)

CryptoPrices/Sources/Data/Sources/NetworkCore/Models/DataPoint.swift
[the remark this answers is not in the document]
minhnimble: Good point @Thieurom, let me omit the public marks on the definition of DataPoint too to make it internally to Repositories. 👍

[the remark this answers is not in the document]
minhnimble: > At the end of the day, what it returns is still the abstract type (CharDataPoint protocol) by casting. So the dependee (GetChartPricesUseCase) doesn't know that. You can omit the public marks on the definition of DataPoint too to make it internally to Repositories.

-> Even though `DataPoint` is a kind of intermediate type for transforming data purposes used specifically on the implementation detail of the `Repositories`, I still think it is better to place it in `NetworkCore` instead of `Repositories`. This is because:
- The current implementation for `DataPoint` is just like what we are doing with the API models like `APICoin` and `APIPrices` as the UseCase also doesn't know about them, but we still make them public for `NetworkExtension`, `DataTestHelpers`, etc. to use, right? Isn't `NetworkCore` the place we should define the **models** for even internal libs of the Data layer to use? 🤔
- It just feels weird that we treat the `DataPoint` class uniquely in a way that `Repositories` can contain other types of classes than repository implementation classes just to limit the access of the `DataPoint` 😅

CryptoPrices/Sources/Data/Sources/NetworkExtension/Parameters/GetChartPricesParameters.swift
minhnimble: > [Updated] Now I realized the `days` parameter may accept a `"max"` value, which may cap the maximum ranges that they can return.  Finally `String` makes sense to me 😄 

Oh okay, the API also supports text constraint, got it now 😂

CryptoPrices/Sources/Data/Sources/NetworkCore/CoinAPIProtocol.swift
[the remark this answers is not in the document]
minhnimble: @Thieurom with the example of different account tiers, I think it starts to make more sense to me now. 👍I think you can just start by explaining with visual examples next time for such architectural ideas like this 🙈

-> I will remove the enum usage for the network layer i.e. `CoinAPIProtocol` 🚀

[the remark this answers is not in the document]
nkhanh44: Thank you, @Thieurom, for the detailed explanation with a great visual example 🙏 and @minhnimble for raising this, which helps me clearer in this architecture 🙏.

CryptoPrices/Sources/Data/Sources/NetworkCore/Models/DataPoint.swift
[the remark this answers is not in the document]
Thieurom: - We don't put the `DataPoint` in the network layer because both interface (`NetworkCore`) and implementation (`NetworkExtension`) don't use it. Its functionality returns `APICoin` and `APIPrices`, which it already includes. But yeah, that's usually fine to define some extra, convenient types for the module consumer 🙏 
- >It just feels weird that we treat the DataPoint class uniquely in a way that Repositories can contain other types of classes than repository implementation classes just to limit the access of the DataPoint

  I don't quite understand your point here 🙏 It means you can't add any new types in the `Repositories` module to support its implementation of the `CoinRepositoryProtocol`? Why is that 🤔 Let's take the `Home` module as an example:
The `Home` module makes a type `HomeView` publicly which conforms to the SwiftUI's `View` protocol. It means it has to have a `body` property with the type of `some View`. In doing so, the `Home` module defines some intermediate types, but for internal use without public them like `WalletStatisticSection`, `TrendingCoinSection`, ... which also just conforms to the `View` protocol. In the end, what `body` returns is `some View` so the consumer doesn't know and doesn't care what exact types it receives under the hood. This is precisely the same situation with the `DataPoint` that conforms to `ChartDataPoint`.

But to conclude, I think putting it in the network layer doesn't hurt, and it has some potential benefits as you mentioned. So let's keep it as is 👍

CryptoPrices/Sources/Data/Sources/NetworkCore/CoinAPIProtocol.swift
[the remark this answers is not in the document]
Thieurom: @minhnimble  @nkhanh44

I would thank and sorry you guys at the same time! I really meant that. Because actually, I don't know all these beforehand. I've been still constantly researching and experimenting while trying to find the best practices to apply to our application.

I think one of the factors that makes it more complex than usual is that we try to fit Clean Architecture into our new architecture. That's the reason why I'd prefer to implement a _pure MVVM_ in the first place 😄

Architecture is always opinionated and debatable. Recently I found a very interesting thread on the Apple Developer Forum that discusses [Stop using MVVM for SwiftUI](https://developer.apple.com/forums/thread/699003) which is of course, still opinionated and debatable 😄 But it has some very interesting points and insights, you guys should take a look at it 🙏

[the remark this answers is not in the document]
minhnimble: Resolved in [bec23cc](https://github.com/nimblehq/nimble-crypto-ios/pull/75/commits/bec23cccdbbed520a37c72692a9dec37d5f77be0) 🎉

CryptoPrices/Sources/Data/Sources/NetworkCore/Models/DataPoint.swift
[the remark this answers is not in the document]
minhnimble: > I don't quite understand your point here 🙏 It means you can't add any new types in the Repositories module to support its implementation of the CoinRepositoryProtocol? Why is that 🤔

-> Ah FYI originally, it is because the name of the library here is `Repositories`, so from my impression, we should only put Repository implementation classes here only, and no model classes like `DataPoint` should be included 🙏. For example:
- `CoinRepository`
- `UserRepository` (for authentication feature)
- `PaymentRepository` (for payment feature)
- So on...

Hope this helps clarify the confusion! 🙏 I see that you have the Home module example there but I think it is a bit off-track of my point 🙈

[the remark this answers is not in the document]
Thieurom: As an updated note, we can still keep the `DataPoint` in the network layer and make it public as is, but we shouldn't attempt to use it on the use case or any higher layer because
>A module of a lower layer depends on one or more modules of the higher layer, but not vice versa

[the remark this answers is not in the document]
minhnimble: Yup, I agree with that idea too 👍

Thieurom (approved):
Everything is great now 🎉 ! Let's 🚀 to 🌕

PR #80, opened 2022-12-30
[#59] [Chore] Use Swift package plugin to support swiftgen generate code for resources

### LINKED ISSUE #59: [Chore] Use Swift package plugin to support `swiftgen` generate code for resources
## Why

Currently, every time when we add new resources (color assets, image assets, fonts, or localized strings) we have to manually run the script `make generate` to update the generated codes.

We have to make this task a build process, specifically, a prebuild phase, so it runs automatically (if needed) when we build the app or the single module that depends on those generated codes.

Xcode 14 (and requires Swift 5.6 and above) supports creating a Swift package plugin that depends on its type, it can be part of the build process or a standalone command that can run directly in Xcode or in the command line.

We'll add a new Swift package plugin that then applies to the module that embeds the resources.

## Who Benefits?

Developers whose tasks require adding new resources.

Thieurom:
- Close #59

## What happened 👀

- Since `swiftgen` has the [official SPM plugin](https://github.com/SwiftGen/SwiftGenPlugin), we'll use it rather than implementing it ourselves (though it's not too hard to do so)
- A SPM plugin is meant to apply to packages individually, so in this PR, we'll add separate `swiftgen.yml` files for packages that are applied rather than using a shared one. It seems strange at first that we have more than one `swiftgen.yml` file in the project, but it makes it more flexible and maintainable, because different packages may have different directory structures, and configurations, ...
- When building the app locally, we might see the validation prompt which in Xcode looks like this:

  207394170-9490e687-e066-4bfa-862c-a4f816b6b43b

  and it's easy to process on the UI.

  However, to run on the CI (headless), it is mandatory to pass the `-skipPackagePluginValidation` flag to the `xcodebuild` command, which I found [here](https://forums.swift.org/t/telling-xcode-14-beta-4-to-trust-build-tool-plugins-programatically/59305) and [here](https://swiftpackageindex.com/lukepistrol/SwiftLintPlugin). So an update for `fastlane`'s `buildAndTest` lane is necessary.

## Insight 📝

- SPM works great so far for the basic tasks and operations, but it needs to evolve more for advanced features, especially in the CI/CD.
- It's great to be able to remove a `homebrew` dependency (`swiftgen`) (by applying a remote SPM plugin) so it makes the automated task run consistently on locals and remote CI.
- We can no longer bother with the generated source files and can be safe to remove the previously-generated ones from version control.

## Proof Of Work 📹

The build log:

Screenshot_2023-01-04_at_15_17_16

The app runs with the correct UI (including fonts, colors, assets, and localized strings):

https://user-images.githubusercontent.com/17972674/210512768-137fa180-f8c8-4a2a-ad4c-e13db1bfe5d6.mp4


nkhanh44 (commented):
Great work @Thieurom ! No more manual steps :D. 

I have a question that since localization is a resource like Color or Image, should we add Home or Mycoin's localization to StyleGuide so that there's no need to add SwiftGenPlugin as a dependency or another swiftgen file to Home or MyCoin? 🙏  Thank you.

Thieurom:
> I have a question that since localization is a resource like Color or Image, should we add Home or Mycoin's localization to StyleGuide so that there's no need to add SwiftGenPlugin as a dependency or another swiftgen file to Home or MyCoin? 🙏 Thank you.

> I have a question that since localization is a resource like Color or Image, should we add Home or Mycoin's localization to StyleGuide so that there's no need to add SwiftGenPlugin as a dependency or another swiftgen file to Home or MyCoin? 🙏 Thank you.

That's a good question @nkhanh44

Actually, although fonts, assets (colors and images) (in Styleguide), and localized strings in (Home and MyCoin) are _resources_ they serve different purposes, and therefore the ways we treat them and process them (with `swiftgen`) are different.
- The resources in Styleguide are meant to use across the app, by means of `import` from many modules. Because it'll expose many UI or styling stuff to make the app looks consistent with the design. That also affects how we config the `swiftgen`: we have to make generated types to be public with `publicAccess: true` in the `swiftgen.yml` so we have something like this:
  ```swift
  public enum Colors {
    public static let bgCurrencyItem = ColorAsset(name: "bg_currency_item")
    public static let bgMain = ColorAsset(name: "bg_main")
    // ...
  }
  ```
- For the localized strings in feature modules (Home and MyCoin): those strings are actually used internally by corresponding package, they are not meant to be public and then used by other packages. Even if both packages may have the same string, let's say `"SeeAll"`, we still should make both packages contain their own versions of that text 🙏 . For this, we don't want to make the generated types to be public. So in their `swiftgen.yml` we won't have `publicAccess: true`, and we have something like this:
  ```swift
  internal enum Strings {
  internal enum Home {
    internal enum MyCoin {
      /// My Coins 😎
      internal static let title = Strings.tr("Localizable", "home.myCoin.title", fallback: "My Coins 😎")
    }
    internal enum SeeAll {
      /// see all ->
      internal static let text = Strings.tr("Localizable", "home.seeAll.text", fallback: "see all ->")
    }
    // ...
  }
  ```

Like I said earlier, even though it seems we have more works to do when distributing large `swiftgen` configurations to some smaller ones but when it comes to modularization I think it's a good way to go. In terms of scale, our app is quite small I still like to think as if we're working on a large one, a super app (like Grab, VinId) 😄 So each module (like Home or MyCoin) is considered an independent, small project. They may be developed by different teams. They can have their own structures, dependencies, and conventions, ... An example is Home may want to use `swiftgen` to generate codes, but `MyCoin` can use another tool like `R.swift` to do the same purpose, etc, ...

Hope that answers your question, and feel free to discuss more 🙏

nkhanh44:
@Thieurom Thank you for the helpful and detailed explanation 🙏 . I got the point :D. It's good to 🚀  now.

nkhanh44 (approved):

minhnimble (approved):
@nkhanh44 I think you have a really nice raise 👍 and thank @Thieurom for the detailed explanation of why we should have the Localization resource per package 🚀 To participate in the discussion, I just have a tiny suggestion for us to consider regarding this:

> For the localized strings in feature modules (Home and MyCoin): those strings are actually used internally by corresponding package, they are not meant to be public and then used by other packages. Even if both packages may have the same string, let's say "SeeAll", we still should make both packages contain their own versions of that text...

-> Having a second thought, I think even our shared package - `Styleguide` can make use of a shared Localization resource, to be honest. Let's imagine in a medium size project, we have around 10+ screens and each screen has the localized string `OK`, then it would be 10+ duplications across the packages in the modular application, right? Isn't the redundancy there quite cumbersome to deal with? And it scales exponentially between the number of strings and the packages that the project has. Hence, it is going to be quite hard to scale well I believe 🙇‍♂️

=> In the end, I am totally not against having each specific localization resource per package, and of course, we need clean separation for sure. It is just that maybe we can consider additionally creating a shared Localization resource in the `Styleguide` for frequently used strings across the application like the example above. Just my two cents 🙏 And yeah if we all think it is a good idea to apply, the implementation can definitely be in a separate branch later on. 🚀

Implementation-wise it LGTM, let's go 🎉

PR #84, opened 2023-01-04
[#44] [UI] As a user, I can see the coin’s statistics on the Detail screen

### LINKED ISSUE #44: [UI] As a user, I can see the coin’s statistics on the Detail screen
## Why

- The user should be able to see the coin’s statistics on the Detail screen.

## Acceptance Criteria

- The user will see the following section:

_In the light mode:_

Market Cap

_In the dark mode:_

Market Cap

There are 3 cells:
**Market Cap:**
- The Market Cap’s value will be based on `market_cap.usd` with the symbol $ in front.
- The Market Cap change percentage’s value will be based on `market_cap_change_percentage_24h`.
  - It's set to `#F15950` with the down indicator when it’s a negative number.
  - It's set to `#10DC78` with the up indicator when it’s a positive number.

**All Time High:**
- The All Time High’s value will be based on `ath.usd` with symbol $ in front.
- The All Time High Change Percentage’s value will be based on `ath_change_percentage.usd`.
  - It's set to `#F15950` with the down indicator when it’s a negative number.
  - It's set to `#10DC78` with the up indicator when it’s a positive number.

**All Time Low:**
- The All Time Low’s value will be based on `atl.usd` with the symbol $ in front.
- The All Time Low Change Percentage’s value will be based on `atl_change_percentage.usd`.
  - It's set to `#F15950` with the down indicator when it’s a negative number.
  - It's set to `#10DC78` with the up indicator when it’s a positive number.

## Design
- https://www.figma.com/file/Iy9eg3ym29pkKGsD5X48O7/Crypto-App-(Community)?node-id=3%3A48&t=xRCVfYvIyS1Y5yrA-0

nkhanh44:
- Close #44

## What happened 👀

- Created UI Coin Statistics in the Detail Screen.
- Add localization for MyCoin.

## Insight 📝

N/A

## Proof Of Work 📹

Dark mode:

https://user-images.githubusercontent.com/25881847/210465417-3ef0affc-816c-4cd5-ad79-5bd99f17db98.mp4

Light mode:

https://user-images.githubusercontent.com/25881847/210465398-33eb977b-7841-4f67-8b57-eb19bb7b5d8c.mp4


CryptoPrices/Sources/MyCoin/Sources/MyCoin/CoinStatisticsSection/CoinStatisticsSection.swift
```diff
@@ -0,0 +1,98 @@
...
+                price: coinDetailItem.ath,
+                isPriceUp: coinDetailItem.isAthChangePercentageUp,
+                percentage: coinDetailItem.athChangePercentage
+            )
+            setUpCoinStatisticsItemView(
+                title: Strings.MyCoin.Atl.title,
+                price: coinDetailItem.atl,
+                isPriceUp: coinDetailItem.isAtlChangePercentageUp,
+                percentage: coinDetailItem.atlChangePercentage
+            )
```
Thieurom, line 32: Just a matter of personal convention, feel free to ignore 🙏

The method `setUpCoinStatisticsItemView` creates and returns a (composite) view based on some inputs. (not too relevant but it's actually a pure function, that returns the same result with the same inputs when running _multiple times_ without side effects). So its name should be `makeCoinStatisticsItemView` or `createCoinStatisticsItemView` to better convey that meaning 🙏

To me, it's even better if we name it as simple as `coinStatisticsItemView` because it's part of another view.

The word `setUp` seems like a thing we mostly do _once_ with a sense that we prepare something or make something to be ready for other things to happen (and as a result usually has some side effects)

CryptoPrices/Sources/MyCoin/Sources/MyCoin/CoinStatisticsSection/CoinStatisticsSection.swift
```diff
@@ -0,0 +1,98 @@
...
+                Text(title)
+                    .foregroundColor(Colors.textMedium.swiftUIColor)
+                    .font(Fonts.Inter.medium.textStyle(.callout))
+                    .frame(height: 22.0)
+                
+                Text(price, format: .dollarCurrency)
+                    .foregroundColor(Colors.titleMedium.swiftUIColor)
+                    .font(Fonts.Inter.medium.textStyle(.callout))
+                    .frame(height: 28.0)
+            }
```
Thieurom, line 63: Please help me confirm the font of the `title` here. It seems that it should have one that differs from the one of `price`, which however these show both use the same `Fonts.Inter.medium.textStyle(.callout)`

Also check again the necessity of setting the frame for these texts (`.frame(height: 22.0)` and `.frame(height: 28.0)`) 🙏

Thieurom (commented):
Just some minor concerns, everything else looks great 👍

CryptoPrices/Sources/MyCoin/Sources/MyCoin/CoinStatisticsSection/CoinStatisticsSection.swift
nkhanh44: Really thank you for the suggestion :D. The `coinStatisticsItemView` sounds more `swiftUI` to me.
I fixed it at https://github.com/nimblehq/nimble-crypto-ios/pull/84/commits/83786ad9d575b07d0f8de4b6c7c0030af3b78667.

CryptoPrices/Sources/MyCoin/Sources/MyCoin/CoinStatisticsSection/CoinStatisticsSection.swift
nkhanh44: Good catch 🙌 . The title should have the `caption` font (size 12.0). I fixed it at https://github.com/nimblehq/nimble-crypto-ios/pull/84/commits/83786ad9d575b07d0f8de4b6c7c0030af3b78667.

>Also check again the necessity of setting the frame for these texts (.frame(height: 22.0) and .frame(height: 28.0)) 🙏

The frame for text is needed here :D. Because in Figma, I don't see the accuracy px space between the title text and price text, as in the following image below:
Screenshot 2023-01-04 at 15 18 59
I set their height frame so they can have their natural padding like the design.

Here is without setting their frame:



Please correct me if I was wrong or misunderstood your point 🙏 .

Thieurom: I think in this case, we can make it simple by adding small vertical padding or providing a custom `spacing` for the `VStack` that contains those texts.
This is the screenshot when I run with `.padding(.vertical, 4.0)` for both texts:
Simulator Screen Shot - iPhone 14 - 2023-01-04 at 16 13 32

It looks almost the same as the POW.

It's up to you to choose which approach, but to me adding padding (or setting `VStack`'s spacing) instead of fixing the frame is less font-dependent 🙏

Thieurom (approved):
🚀

CryptoPrices/Sources/MyCoin/Sources/MyCoin/CoinStatisticsSection/CoinStatisticsSection.swift
```diff
@@ -0,0 +1,98 @@
...
+            VStack(alignment: .leading) {
+                Text(title)
+                    .foregroundColor(Colors.textMedium.swiftUIColor)
+                    .font(Fonts.Inter.medium.textStyle(.caption))
+                    .padding(.vertical, 4.0)
+                
+                Text(price, format: .dollarCurrency)
+                    .foregroundColor(Colors.titleMedium.swiftUIColor)
+                    .font(Fonts.Inter.medium.textStyle(.callout))
+                    .padding(.vertical, 4.0)
```
minhnimble, line 62: Just curious, why don't we define a fix spacing value directly into the `VStack` but we have to add the vertical padding individually like this? 🤔

minhnimble (commented):
Just a small concern left from me, the rest LGTM 🚀

CryptoPrices/Sources/MyCoin/Sources/MyCoin/CoinStatisticsSection/CoinStatisticsSection.swift
nkhanh44: Thank you for the question 🙏 .
I had no specific purpose or tradeoff for adding the vertical padding individually per `Text` view instead of defining a fixed spacing value directly into the `VStack` , except for setting padding for it :D. IMO both approaches are fine for me.

Setting spacing value directly into `VStack` may reduce the repeated code 👍, a nice suggestion. Thank you for the raise 🙇 . I fixed it at https://github.com/nimblehq/nimble-crypto-ios/pull/84/commits/12ce15bee90af12af289e290ccc5eba464e95c1d.

minhnimble (approved):
Let's go 🚀

PR #88, opened 2023-01-06
[#62] [Integrate] As a user, I can pull to refresh data on the Home screen

### LINKED ISSUE #62: [Integrate] As a user, I can pull to refresh data on the Home screen
## Why

- The user should be able to refresh all data on the home screen by pulling.

## Acceptance Criteria

When pulling, the spinner will be displayed as the following:

_In the light mode:_

Screenshot 2022-12-20 at 13 14 37

_In the dark mode:_

Screenshot 2022-12-20 at 13 14 42

We can use the default iOS spinner.

The following APIs will be called: `GET /coins/markets` for the `trending coins` and `my coins` sections with the initial parameters.

## Resources
-  [Figma design](https://www.figma.com/file/Iy9eg3ym29pkKGsD5X48O7/Crypto-App-(Community)?node-id=0%3A1&t=NhL8SGKxuUJWFmlw-0)

Thieurom:
- Close #62

## What happened 👀

Provide a description of the **changes** this pull request brings to the codebase. Additionally, when the pull request is still being worked on, a checklist of the planned changes is welcome to track progress.

## Insight 📝

Describe in detail why this solution is the most appropriate, which solution you tried but did not go with, and how to test the changes. References to relevant documentation are welcome as well.

## Proof Of Work 📹

Show us the implementation: screenshots, GIFs, etc.


CryptoPrices/Sources/Home/Sources/Home/Resources/en.lproj/Localizable.strings
```diff
@@ -1,4 +1,4 @@
-"home.myCoin.title" = "My Coins 😎";
+"home.myCoins.title" = "My Coins 😎";
```
minhnimble, line 1: 👍

CryptoPrices/Sources/Home/Sources/Home/CommonViews/Header.swift
```diff
@@ -0,0 +1,32 @@
...
+//  Header.swift
+//  Home
+//
+//  Created by Doan Thieu on 05/01/2023.
+//
+
+import Styleguide
+import SwiftUI
+
+struct Header: View {
```
minhnimble, line 11: Do you think we should be explicit about the Header name here that we should turn into `HomeHeader` instead? As this header layout looks kinda specific for the Home screen and we might have different type of headers in the future. What do you think? 🤔

CryptoPrices/Sources/Styleguide/Sources/Styleguide/View+Extensions/View+RenderIf.swift
```diff
@@ -0,0 +1,30 @@
+//
+//  View+RenderIf.swift
+//  Styleguide
+//
+//  Created by Doan Thieu on 05/01/2023.
+//
+
+import SwiftUI
+
+private struct RenderIfModifier: ViewModifier {
```
minhnimble, line 10: Nice extension 🚀

minhnimble (approved):
The implementation is really nice! 🎉

As a side note, can you also have a record case that first opening the application with no internet, then trigger a reload with internet connection to showcase the new `renderIf` call for the P.o.W? That would be a great demonstration 🙏

CryptoPrices/Sources/Home/Sources/Home/Resources/en.lproj/Localizable.strings
Thieurom: 🙏

CryptoPrices/Sources/Styleguide/Sources/Styleguide/View+Extensions/View+RenderIf.swift
Thieurom: Thank you 🙏

CryptoPrices/Sources/Home/Sources/Home/CommonViews/Header.swift
Thieurom: Yes, I totally agree 👍 I'll make it updated!

CryptoPrices/Sources/Styleguide/Sources/Styleguide/View+Extensions/View+RenderIf.swift
```diff
@@ -0,0 +1,30 @@
...
+
+    let condition: Bool
```
nkhanh44, line 12: As we have the initializer below, we should set this private access 🙏 
```suggestion
   private let condition: Bool
```

CryptoPrices/Sources/Home/Tests/HomeTests/HomeViewModelSpec.swift
```diff
@@ -120,6 +120,79 @@ final class HomeViewModelSpec: QuickSpec {
...
+                    it("gets the correct value for trendingCoins") {
+                        await expect { await homeViewModel.trendingCoins }
+                            .to(equal(expectedTrendinglCoins))
+                    }
+                }
+            }
+
+            describe("its fetchAllData() call") {
+
+                context("when myCoinsUseCase returns success, trendingCoinsUseCase returns failure") {
```
nkhanh44, line 151: I think we can omit this `describe` and the below one and put all the contexts into the one `describe("its fetchAllData()) `call") above. What do you think?

nkhanh44 (approved):
Great work 👍 , just a few suggestions:D.

CryptoPrices/Sources/Styleguide/Sources/Styleguide/View+Extensions/View+RenderIf.swift
Thieurom: Thank you for spotting out 🙏 , updated with https://github.com/nimblehq/nimble-crypto-ios/pull/88/commits/fa7584af1a0ab5e480fac76ff19766889e7c0081

CryptoPrices/Sources/Home/Tests/HomeTests/HomeViewModelSpec.swift
Thieurom: Oh, that's my bad 🙏  Updated in https://github.com/nimblehq/nimble-crypto-ios/pull/88/commits/6873240d834d3f2e0f2a6fde4cb33bf9f17fe8f1

Thieurom:
> As a side note, can you also have a record case that first opening the application with no internet, then trigger a reload with internet connection to showcase the new `renderIf` call for the P.o.W? That would be a great demonstration 🙏

A PoW video for the case of launching the app without an internet connection was added 🙏

PR #92, opened 2023-01-09
[#59] [Chore] Use Swift package plugin to support swiftgen generate code for resources - part 2

### LINKED ISSUE #59: [Chore] Use Swift package plugin to support `swiftgen` generate code for resources

Thieurom:
- Close #59

## What happened 👀

We did set the `-skipPackagePluginValidation` flag when building and testing with `fastlane` in the PR #80.
This PR did the same when building and archiving (to deploy to Firebase)

## Insight 📝

`N/A`

## Proof Of Work 📹

`N/A`

### DIFF
--- a/CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
+++ b/CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
@@ -4,6 +4,7 @@
 //  Created by Mike Pham on 04/12/2022.
 //
 
+import DomainTestHelpers
 import Styleguide
 import SwiftUI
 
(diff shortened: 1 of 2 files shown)


minhnimble (commented):
The changes LGTM 🚀 

By the way, there seems to be an importing issue when archiving that it is unable to recognize `MockCoinDetail` which could make the build fail (this won't happen on debug builds) and we will need to move the `import DomainTestHelpers` out of the `#if DEBUG` marcro. -> I think we can do that in this PR too for fixing another potential release issue. 🙏

Screenshot 2023-01-09 at 20 28 01

minhnimble (approved):

Issue #63, opened 2022-12-20
[Integrate] As a user, I can pull to refresh data on the Detail screen


nkhanh44:
## Why

- The user should be able to refresh all data on the detail screen by pulling.

## Acceptance Criteria

When pulling, the spinner will be displayed as the following:

_In the light mode:_

![Screenshot 2022-12-20 at 13 14 50](https://user-images.githubusercontent.com/25881847/208603692-a54f9c92-e748-45da-badc-9187feb09b2d.png)

_In the dark mode:_

![Screenshot 2022-12-20 at 13 14 55](https://user-images.githubusercontent.com/25881847/208603706-df3d2332-1381-45eb-8582-0a18d4366c89.png)

We can use the default iOS spinner.

The following APIs will be called: 

- Get coin detail: `GET /coins/{id}`
- Get coin prices:
    - For `1D, 7D, 1M, 1Y 5Y` price: `GET /coins/{id}/market_chart`

The chart will be displayed with `1D` timeframe, and the timeframe is set to 1D as default.

## Resources
-  [Figma design](https://www.figma.com/file/Iy9eg3ym29pkKGsD5X48O7/Crypto-App-(Community)?node-id=0%3A1&t=NhL8SGKxuUJWFmlw-0)

PR #93, opened 2023-01-10
[#46] [UI] As a user, I can see the Buy & Sell buttons on the Detail screen

### LINKED ISSUE #46: [UI] As a user, I can see the Buy & Sell buttons on the Detail screen
## Why

- The user should be able to see the Buy & Sell buttons at the bottom on the Detail screen.

## Acceptance Criteria

- The user will see the following Buy & Sell buttons:

_In light mode:_

Market Cap

_In dark mode:_

Market Cap

Note: We only handle the UI, not Integrate or backend for this UI.

Thieurom:
- Close #46

## What happened 👀

- Refactor `MyCoinView`
  - Replace `ScrollView` with `List` (this is also necessary when we implement the [pull-to-refresh](https://github.com/nimblehq/nimble-crypto-ios/issues/63) later)
  - Move `Section` into `Styleguide` and make it public so it can be used by `HomeView` and `MyCoinView`
- Create 2 button styles `PrimaryButtonStyle` and `SecondaryButtonStyle`, applied for __Buy__ and __Sell__ buttons respectively for the `MyCoinView`

## Insight 📝

When it comes to styling the `Button` (or any SwiftUI views in general), there's more than one way. I once talked about it in our chapter retro, and had some examples of that in our [swiftui-showcases](https://github.com/nimblehq/swiftui-showcases/pull/8/files#diff-f0a542b787c9bd27040c2eb02ce2e61d2cc09e78e341e17fafc33a3201032b0e). But IMHO use the `buttonStyle` seems the most idiomatic one 🙏 

## Proof Of Work 📹

|iPhone 14 - Dark Mode|iPhone SE - Light Mode|
|---|---|
|<video src="https://user-images.githubusercontent.com/17972674/211490356-3b826ac3-318d-4b0e-b5d1-aeee721cf4e9.mp4">|<video src="https://user-images.githubusercontent.com/17972674/211491023-94671c48-51e4-48a8-91c7-f60ea16b0194.mp4">|

### DIFF
--- a/CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
+++ b/CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
@@ -8,28 +8,39 @@ import DomainTestHelpers
 import Styleguide
 import SwiftUI
 
+typealias Section = Styleguide.Section
+
 public struct MyCoinView: View {
 
     @EnvironmentObject var myCoinState: MyCoinState
 
     public var body: some View {
-        contentView
-            .navigationBarBackButtonHidden()
-            .navigationBarItems(leading: backButton, trailing: likeButton)
-            .navigationTitle(viewModel.coinDetail?.name ?? "")
-            .navigationBarTitleDisplayMode(.inline)
-            .gesture(
-                DragGesture()
-                    .onEnded { value in
-                        // Handle swipe left-to-right
-                        if value.startLocation.x < 20.0 && value.translation.width > 50.0 {
-                            myCoinState.didSelectBack = true
-                        }
+        List {
+            currentPriceSection
+            priceChartSection
+            coinStatisticsSection
+        }
+        .listStyle(.plain)
+        .background(Colors.bgMain.swiftUIColor)
+        .navigationBarBackButtonHidden()
+        .navigationBarItems(leading: backButton, trailing: likeButton)
+        .navigationTitle(viewModel.coinDetail?.name ?? "")
+        .navigationBarTitleDisplayMode(.inline)
+        .gesture(
+            DragGesture()
+                .onEnded { value in
+                    // Handle swipe left-to-right
+                    if value.startLocation.x < 20.0 && value.translation.width > 50.0 {
+                        myCoinState.didSelectBack = true
                     }
-            )
-            .task {
-                await viewModel.fetchCoinDetail(id: myCoinState.id)
-            }
+                }
+        )
+        .safeAreaInset(edge: .bottom, content: {
+            footerView
+        })
+        .task {
+            await viewModel.fetchCoinDetail(id: myCoinState.id)
+        }
     }
 
     @ObservedObject private var viewModel: MyCoinViewModel
@@ -41,24 +52,29 @@ public struct MyCoinView: View {
 
 private extension MyCoinView {
 
-    var contentView: some View {
-        ScrollView {
+    var currentPriceSection: some View {
+        Section {
             CurrentPriceSection()
-            Spacer(minLength: 38.0)
+        }
+    }
+
+    var priceChartSection: some View {
+        Section {
             PriceLineChartSection()
                 .frame(height: 196.0)
-            // TODO: - Add time filter here
-            // TODO: - Remove dummy
+
             TimeFrameSection()
+        }
+    }
+
+    var coinStatisticsSection: some View {
+        Section {
             CoinStatisticsSection(
                 coinDetailItem: CoinDetailItem(
                     coinDetail: MockCoinDetail.single
                 )
             )
         }
-        .clipped(antialiased: false)
-        .frame(maxHeight: .infinity)
-        .background(Colors.bgMain.swiftUIColor)
     }
 
     var backButton: some View {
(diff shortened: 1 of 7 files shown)


CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
```diff
@@ -8,28 +8,39 @@ import DomainTestHelpers
 import Styleguide
 import SwiftUI
 
+typealias Section = Styleguide.Section
```
minhnimble, line 11: This looks like a repeated `typealias` declaration, I think we can define one in the `Styleguide` module directly and just use it here and for `HomeView` also. 🙏

minhnimble (commented):
Just a tiny suggestion, rest LGTM 🚀

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
Thieurom: The `Section` belongs to the `Styleguide`, so if we want to define a `typealias` in the `Styleguide`, it must have a different name like `ListSection`, `CustomSection`, or something like that. But it doesn't make sense, we can just name it `struct ListSection` or `struct CustomSection`, so no `typealias` is necessary in the first place 🙏 

To be honest, those names are quite redundant and too little too specific. For e.g. let take the `ListSection` name as an example. `Section` can be used in any container view like `VStack`, `HStack` or even nested in another `Section`, not only in the `List`. So the name `ListSection` is not only too specific but also not totally correct 🙏 

I'd prefer to keep the original name as `Section`. And it's up to the modules which use the `Section` (`Home` and `MyCoin`) to define whatever `typealias` they want to be to avoid ambiguity with `SwiftUI`'s `Section`. For e.g in the `Home` it could be `typealias HomeSection = Styleguide.Section`, while in the `MyCoin` it could be `typealias MyCoinSection = Styleguide.Section`.

So instead of seeing it as the repetitive, another way to think about it is that, it just happens (coincidentally 😅) that 2 module `Home` and `MyCoin` define their own `typealias` with the same name `typealias Section = Styleguide.Section`. What do you think 🙏 ?

minhnimble: Ohh you are right that we don't need a typealias anymore if it is placed in the `Styleguide` package, my bad 🙇‍♂️ Yep for sure there is not a need for renaming them to `HomeSection` or `MyCoinSection`, keeping just `Section` is fine as it is referring to the same struct in our common views 👍

Now, since the typealias is only used one time e.g. in [MyCoinView.swift](https://github.com/nimblehq/nimble-crypto-ios/pull/93/files/8f75a4712ac5c779545a99e5422c1a75c970746a#diff-ac6dd79a4e1836b63d9a2b484a757a66804ab87e1271482f22d347a2cf06d1d2) for `MyCoin` package, what do you think about turning it to private instead? Later on, when there is a new need for using it in multiple places, it would be better to move it out of the `MyCoinView` class and place it in a new file for shared usage 😁

CryptoPrices/Sources/Styleguide/Sources/Styleguide/ButtonStyles/SecondaryButtonStyle.swift
```diff
@@ -0,0 +1,28 @@
...
+//
+//  Created by Doan Thieu on 10/01/2023.
+//
+
+import SwiftUI
+
+public struct SecondaryButtonStyle: ButtonStyle {
+
+    private static let backgroundColor = Colors.carnation.swiftUIColor
+    private static let selectedBackgroundColor = Colors.carnation.swiftUIColor.opacity(0.5)
```
nkhanh44, line 13: Do we have any other purpose for using `static let`, except for defining it as a constant? Maybe only `let` is enough? Thank you 🙏 .

nkhanh44 (commented):
Good work 👍 I have a small concern, please help me to clarify 🙏 .

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
Thieurom: @minhnimble Yes, totally agree 👍
Will make both `typealias`es in the `Home` and `MyCoin` private!

CryptoPrices/Sources/Styleguide/Sources/Styleguide/ButtonStyles/SecondaryButtonStyle.swift
Thieurom: @nkhanh44  That's a good question 🙏

The `static let` make these properties _type property_, not _instance property_. So regardless of how many instances of `SecondaryButtonStyle` we create, these properties are created once, and we'll save a little memory here since the compiler won't have to allocate memories on the stack for these for every instance.

This is usually used when we want to share data between instances, especially when the data is expensive to create. In Swift, `static` also makes data lazily initialized. Let's say we want to add a `DateFormatter` (not make sense here, just an example):
```swift
public struct SecondaryButtonStyle: ButtonStyle {

  private static var dateFormatter: DateFormatter = {
    let formatter = DateFormatter()
    formatter.dateStyle = .medium
    // ... other config
    return formatter
  }()
}
```

In any _instance method_, we can access through `SecondaryButtonStyle.dateFormatter` or the preferred way is `Self. dateFormatter` (with the uppercased `S`). (In _static method_, just use `dateFormatter` directly). Hope this clears your concern 🙏

nkhanh44: Thank you  🙏. It's clear to me.

nkhanh44 (approved):
🚀

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
Thieurom: Updated in https://github.com/nimblehq/nimble-crypto-ios/pull/93/commits/13771ebb808b2a92d7319399586989f1f2a8472b 🙏

minhnimble (approved):
LGTM now, let's go 🚀

PR #94, opened 2023-01-10
[#38] [Integrate] As a user, I can see the chart on the Detail screen

### LINKED ISSUE #38: [Integrate] As a user, I can see the chart on the Detail screen
## Why

- The user should be able to see the chart on the Detail screen.

## Acceptance Criteria

- On the Detail screen, the following chart can be seen:

_In light mode:_

_In dark mode:_

For both the light and dark modes, we have:
- The line chart is a linear line with the color `#00BDB0`.
- The highest price text in that frame has the color `#10DC78`.

Screenshot 2022-12-15 at 23 01 48

- The lowest price text in that frame has the color `#F15950`

Screenshot 2022-12-15 at 23 01 56

- In a timeframe, the latest data (price) will be on the right, and the oldest data will be on the left.

Screenshot 2022-12-19 at 22 55 48

**On behavior:**
- The chart will set fit and **no zooming** for the chart with any gestures.
- No animation is needed when the chart is displayed.

_Note: How the detailed range time is filtered in days, months, and years_ from [API document](https://www.coingecko.com/en/api/documentation):
```
Data granularity is automatic (cannot be adjusted)
1 day from current time = 5 minute interval data
1 - 90 days from current time = hourly data
above 90 days from current time = daily data (00:00 UTC)
```

```JSON
{
  "prices": [
    [
      1669284289704,
      16595.246079387816
    ],
    [
      1669284585059,
      16602.076684659474
    ],
    [
      1669284921814,
      16606.11839794301
    ],
    [
      1669285188016,
      16606.20441243831
}
```

## Design
https://www.figma.com/file/Iy9eg3ym29pkKGsD5X48O7/Crypto-App-(Community)?node-id=3%3A107&t=PLQA07bsooNZNzKn-0

## Resources

https://github.com/danielgindi/Charts

- From the design and the lib above, `LineChart (gradient fill)` in Charts lib is the most identity to the design.

Screenshot 2022-12-20 at 00 25 24

minhnimble:
- Close #38

## What happened 👀

- Remove all hard-coded data.
- Integrate `GetChartPricesUseCase` and its interface into `MyCoinViewModel` for calling actual API to get the data.
- Update the `PriceLineChartView` to support showing localized `No Data` text when there is no data from the server.
- Add new unit tests for `MyCoinViewModel` to reflect the corresponding logic.

## Insight 📝

- Currently, the extensions in the `Styleguide` package are a bit messed up and all over the places so I did a bit of grouping refactor to make them look more organized. Same thing applied for the Mocking structs in the `DomainTestHelpers`.
- When testing getting charts data from the APIs, there can be cases that it reached the allowed limits (error code 429) and in such cases, we won't be able to retrieve new data for testing purposes. Therefore, please test the API wisely. 🙏
  

**P/S:** The limit seems to be removed per day. So when the limit is reached, we can resume testing in the next day 🏃

## Proof Of Work 📹

The chart is now populated with actual data from the server:

https://user-images.githubusercontent.com/70877098/211712403-7d2311c5-7aed-4579-9b3f-7a9724df773c.MP4

In case there is no data, the chart show `No data` text accordingly:

https://user-images.githubusercontent.com/70877098/211759100-f54779a3-3272-4ef1-8cb9-0b76fdad9b0b.MP4

All new unit tests passed 🚀

Screenshot 2023-01-12 at 14 12 38


CryptoPrices/Sources/MyCoin/Sources/MyCoin/ChartSectionView/PriceLineChartView.swift
```diff
@@ -61,6 +60,13 @@ struct PriceLineChartView: UIViewRepresentable {
         }
         dataSet.drawFilledEnabled = true
     }
+
+    private func formatNoDataText(uiView: LineChartView) {
+        uiView.noDataText = Strings.MyCoin.PriceLineChart.noDataText
+        uiView.noDataFont = Fonts.Inter.medium.font(size: 14.0)
+        uiView.noDataTextColor = Colors.textMedium.color
+        uiView.noDataTextAlignment = .center
+    }
```
Thieurom, line 69: Regarding how we implement `PriceLineChartView` which conforms to `UIViewRepresentable`, I have some suggestions. Please consider checking this, and hope you find it helpful 🙏 

According to [Apple documentation](https://developer.apple.com/documentation/swiftui/uiviewrepresentable/makeuiview(context:)):
> The system calls this method [[makeUIView(context:)](https://developer.apple.com/documentation/swiftui/uiviewrepresentable/makeuiview(context:))] only once, when it creates your view for the first time. For all subsequent updates, the system calls the [updateUIView(_:context:)](https://developer.apple.com/documentation/swiftui/uiviewrepresentable/updateuiview(_:context:)) method.

So there're some points worth noting:
- We don't have to keep an instance of `LineChartView` around as a stored property (`let lineChart = LineChartView()`. We just create one and return it in the `makeUIView(context:)`. When SwiftUI wants to make an update, it hands us this exact instance as the parameter in `updateUIView(_ uiView: LineChartView, context: Context)`.
- Any view configurations are meant to be set once, we should place them in the `makeUIView(context:)` together with creating the view. Any view configurations that require change according to the latest state and data should be placed in the `updateUIView(_:context:)` while making the view update.
  
  The exception I believe is that if the one-shot configuration (set once) is expensive with maybe a heavy computation, we can delay by placing it in the `updateUIView(_:context:)` because we want to return the view as quickly as possible for the users. The SwiftUI always calls this method at least once right after `makeUIView(context:)` even if the initial state doesn't change. If that'd be a case, we're still able to set it once by simply adding a `Bool` flag.

Therefore, the _core definition_ of `PriceLineChartView` can be rewritten with small adjustments and arrangements as follows 🙏 
```swift
struct PriceLineChartView: UIViewRepresentable {

    let entries: [ChartDataEntry]

    func makeUIView(context: Context) -> LineChartView {
        let lineChart = LineChartView()

        lineChart.rightAxis.enabled = false
        lineChart.leftAxis.enabled = false
        lineChart.xAxis.enabled = false
        lineChart.legend.enabled = false
        lineChart.minOffset = 0.0
        lineChart.isUserInteractionEnabled = false
        lineChart.noDataText = Strings.MyCoin.PriceLineChart.noDataText
        lineChart.noDataFont = Fonts.Inter.medium.font(size: 14.0)
        lineChart.noDataTextColor = Colors.textMedium.color
        lineChart.noDataTextAlignment = .center
        lineChart.delegate = context.coordinator

        return lineChart
    }

    func updateUIView(_ uiView: LineChartView, context: Context) {
        defer {
            uiView.notifyDataSetChanged()
        }

        guard !entries.isEmpty else {
            uiView.data = nil
            return
        }

        let dataSet = LineChartDataSet(entries: entries)
        formatDataSet(dataSet: dataSet)
        uiView.data = LineChartData(dataSet: dataSet)
    }

    // other methods
    // ...
}
```

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
```diff
@@ -40,6 +40,7 @@ public struct MyCoinView: View {
         })
         .task {
             await viewModel.fetchCoinDetail(id: myCoinState.id)
+            await viewModel.fetchChartPricesData(id: myCoinState.id)
```
Thieurom, line 43: Please help to re-arrange so these method calls can run in parallel. You can refer to what we've done for the `Home` [here](https://github.com/nimblehq/nimble-crypto-ios/blob/a1b4ed874a0d8900d89108df54bb4e0dd068fe18/CryptoPrices/Sources/Home/Sources/Home/HomeView/HomeViewModel.swift#L59)

Thieurom (commented):
Great implementation and neat re-organizing 🎉 
I had some small suggestions, hope you find them helpful 🙏

CryptoPrices/Sources/MyCoin/Sources/MyCoin/ChartSectionView/PriceLineChartView.swift
```diff
@@ -22,16 +22,15 @@ struct PriceLineChartView: UIViewRepresentable {
...
-        // TODO: Localize this text for no data case in the integration task
-        uiView.noDataText = "No Data"
-        uiView.data = LineChartData(dataSet: dataSet)
+        uiView.data = entries.isEmpty ? nil : LineChartData(dataSet: dataSet)
         uiView.rightAxis.enabled = false
         uiView.leftAxis.enabled = false
         uiView.xAxis.enabled = false
         uiView.legend.enabled = false
         uiView.minOffset = 0.0
         uiView.isUserInteractionEnabled = false
```
nkhanh44, line 31: Do you think for this configuration we should put them in the `makeUIView` function because we want to set them once instead of updating along with view changes? please correct me 🙏 .

nkhanh44 (commented):
LGTM 👍 .Just a tiny suggestion outside the scope that I overlooked before.

CryptoPrices/Sources/MyCoin/Sources/MyCoin/ChartSectionView/PriceLineChartView.swift
minhnimble: that is definitely a valid point, I can update it in this PR too, thanks for the raise 🙏

CryptoPrices/Sources/MyCoin/Sources/MyCoin/ChartSectionView/PriceLineChartView.swift
minhnimble: Your suggestion is definitely a good improvement, thanks for the explanation! I will update it 🚀

minhnimble: Regarding the guard check, it might not really be necessary and `if else` check should be cleaner here instead 🙏

nkhanh44 (approved):
🚀

Thieurom (approved):
🚀

PR #95, opened 2023-01-11
[Chore] Add KIF and KIF Target

blyscuit:
- Close #

## What happened 👀

Provide a description of the **changes** this pull request brings to the codebase. Additionally, when the pull request is still being worked on, a checklist of the planned changes is welcome to track progress.

## Insight 📝

Describe in detail why this solution is the most appropriate, which solution you tried but did not go with, and how to test the changes. References to relevant documentation are welcome as well.

## Proof Of Work 📹

Show us the implementation: screenshots, GIFs, etc.


CryptoPricesKIFUITests/Sources/Specs/HomeSpec.swift
```diff
@@ -0,0 +1,28 @@
...
+//  CryptoPricesKIFUITests
+//
+//  Created by Bliss on 11/1/23.
+//
+
+@testable import Home
+import Nimble
+import Quick
+
+final class HomeSpec: QuickSpec {
```
minhnimble, line 12: It looks like we are putting this KIF test in the main project, but since we are using a modular architecture, this UI test should be placed directly in the `Home` package instead. What do you guys think? @Thieurom @nkhanh44

Thieurom: Firstly, welcome to CryptoPrices @blyscuit 🎉 

@minhnimble Yes, that's exactly what I was going to say 😄 

For our architecture, the tests of a module will be placed in the corresponding package. So basically we would have two new test targets for the `Home` and `MyCoin` respectively.

The `CryptoPricesKIFUITests` target you created is still useful if we want to make some kind of _integration test_, like tapping a row on the TrendingCoins section on the Home will navigate to the MyCoin with correct UIs, ...

I think today I'm gonna take a look at [KIF](https://github.com/kif-framework/KIF) to check how to best integrate it into our project 🙏

nkhanh44: A warm welcome to @blyscuit 🕺 .
@minhnimble @Thieurom I agree 💪 . I used to use KIF before but in our architecture, it's something I need to research a little bit :D. I will take a look at this after my tasks are done 🙏 .

blyscuit: KIF is "an iOS integration test framework", it requires some sort of 'app' to run the test. Underneath it works similarly to XCUITest (meant as a replacement of XCUITest): Find view in app > interact with app via accessibility components.

I think this discussion is referring to Unit test of the views, which can be done through [ViewInspector](https://github.com/nalexn/ViewInspector) that David presented or with [Snapshot Testing](https://github.com/pointfreeco/swift-snapshot-testing) that Thieu briefly discussed.

blyscuit: The reason I'm pushing `KIF` to the chapter and this project is because iOS dev are less likely to want to add integration test because how fragile `XCUITest` is. But with `KIF` we can be confident about our integration test and is way easier to work with.

blyscuit: On the modular architecture note, I think it will make sense to add `KIF` to the module if the module act as a user's story ie: register, kyc, feature1, feature2. Currently this modular architecture is separating the feature module to smaller UI modules.

Thieurom: Thank you for your insights @blyscuit 🙏

> KIF is "[an iOS integration test framework](https://github.com/kif-framework/KIF#kif-ios-integration-testing-framework)", it requires some sort of 'app' to run the test.

That's what I thought in the first place
> The CryptoPricesKIFUITests target you created is still useful if we want to make some kind of integration test, like tapping a row on the TrendingCoins section on the Home will navigate to the MyCoin with correct UIs, ...

So basically, with the same `KIF` framework (without an additional one like `ViewInspector`) we can achieve 2 purposes:
- Add the _Unit test of the views_ for each module. It needs some extra work like wrap in a `UIHostingController` (it can be extracted to a helper/extension though) among other things. I think we don't have time for this now!
- Add the _integration tests_, which need to be implemented on the app level like you did here. From what I've seen, it's quite similar with another tool like Appium (I remembered my QA guys and gals wrote alot of test scripts, scenarios in my former companies with this 😄). We should keep going with this! Still I need sometime to play around with `KIF` before making some feedback 🙏

Thieurom:
Hi @blyscuit

I checked out this branch and played around with it a little bit. There're some concerns I'd like to share here; kindly clarify 🙏 I might miss something since I didn't have too much time with `KIF.`

- We kind of agree that the test for individual _feature view_ should be placed in the corresponding package; the `CryptoPricesKIFUITests` target was meant for the _integration test_. So the newly-added test class (and its filename) should be `AppSpec,` and it would start with something like this:
  ```swift
  @testable import Home
  @testable import MyCoin
  import Nimble
  import Quick

  final class AppSpec: QuickSpec {

    override func spec() {
      //
    }
  }
  ```
- When I ran the test for `HomeSpec` locally, the test passed, but the logs showed many `Main Thread Checker` issues. It seems to do with the `KIF`'s internal implementation (how it _launch_ and _run_ the app.

  Screenshot 2023-01-16 at 09 59 43
  Also, the logs show the app called the API. I think we might need some setup to inject the mocked dependencies for the app before executing any tests 🙏

- When I ran the test locally but with a different expected result (I'd like to see how `KIF` fails the test):
  ```swift
  self.tester().waitForView(
    withAccessibilityLabel: "Lorem Ipsum"
  )
  ```
  It ran quite slowly and logged hundreds of `[Assert] Unsupported enumeration of UIWindowScene windows on non-main thread.` then crashed:

  https://user-images.githubusercontent.com/17972674/212593106-741cedc0-5315-45bb-9beb-1564d9e8edda.mov

- The test ran on the CI (GitHub Action) slowly and failed; strangely, it didn't fail the workflow 🤔 
  Screenshot 2023-01-16 at 10 10 57
  I guess maybe the test still passed, but then there's something wrong when the `fastlane` try to write the log or test report to the filesystem.

CryptoPricesKIFUITests/Sources/Specs/HomeSpec.swift
blyscuit: Your two points here are correct.

> It needs some extra work like wrapping in a `UIHostingController`

With the current structure of the modules (`Home` and `MyCoin`), the UIHostingController would also be implemented in a new Target (same or new project file) because the Swift module has limitation of creating an app target inside the module.

> it's quite similar to another tool like Appium

Same as Appium, but much faster build & run and written in Obj-c/Swift.

blyscuit:
@Thieurom 
> So the newly-added test class (and its filename) should be AppSpec

What is the plan when we include more tests to a growing app? Having one AppSpec would be too bloated? What do you think if we name the test specs by the name of the feature flow for example `CoinDetailSpec`, `LoginSpec`, etc?

> [Assert] Unsupported enumeration of UIWindowScene windows on non-main thread. then crashed

This is the same issue with 

> the logs showed many Main Thread Checker issues.

I'm taking a look at it. Could be because `parallel testing` 🤔.

 > CI (GitHub Action) slowly and failed

I will try with a failing test on the CICD of this project.

blyscuit:
@Thieurom `Main Thread` error comes from Quick(6.x) and Nimble(11.x) newest versions. This error does not happen in when running KIF with XCTest or Quick(5.x). To solve this we add `@MainActor` to Quick spec
```

    override func spec() {

        describe("a Home screen") {

            describe("its open") {

                it("it shows its ui components") {  @MainActor in
                    self.tester().waitForView(
                        withAccessibilityLabel: Strings.Home.Title.text
                    )
                }
            }
        }
    }
```
but this will require adding MainActor to all KIF functions. Alternatively we can 
```

    override func spec() {

        describe("a Home screen") { @MainActor in

            describe("its open") {

                it("it shows its ui components") {
                    self.tester().waitForView(
                        withAccessibilityLabel: Strings.Home.Title.text
                    )
                }
            }
        }
    }
```
but it will result in the [warning](https://github.com/apple/swift/issues/61607).
Screen Shot 2023-01-16 at 14 58 14
Personally I prefer the last way but devs will have to ignore the error 😭 .

I leave this to the team to decide 🙏

PR #96, opened 2023-01-11
[#45] [Integrate] As a user, I can see the coin’s statistics on the Detail screen

### LINKED ISSUE #45: [Integrate] As a user, I can see the coin’s statistics on the Detail screen
## Why

- The user should be able to see the coin’s statistics on the Detail screen.

## Acceptance Criteria

- The user will see the following section:

_In the light mode:_

Market Cap

 _In the dark mode:_

Market Cap

There are 3 cells:
**Market Cap:**
- The Market Cap’s value will be based on `market_cap.usd` with the symbol $ in front.
- The Market Cap change percentage’s value will be based on `market_cap_change_percentage_24h`.
  - It's set to `#F15950` with the down indicator when it’s a negative number.
  - It's set to `#10DC78` with the up indicator when it’s a positive number.

**All Time High:**
- The All Time High’s value will be based on `ath.usd` with symbol $ in front.
- The All Time High Change Percentage’s value will be based on `ath_change_percentage.usd`.
  - It's set to `#F15950` with the down indicator when it’s a negative number.
  - It's set to `#10DC78` with the up indicator when it’s a positive number.

**All Time Low:**
- The All Time Low’s value will be based on `atl.usd` with the symbol $ in front.
- The All Time Low Change Percentage’s value will be based on `atl_change_percentage.usd`.
  - It's set to `#F15950` with the down indicator when it’s a negative number.
  - It's set to `#10DC78` with the up indicator when it’s a positive number.

API response from #12 .

## Design
- https://www.figma.com/file/Iy9eg3ym29pkKGsD5X48O7/Crypto-App-(Community)?node-id=3%3A48&t=xRCVfYvIyS1Y5yrA-0

nkhanh44:
- Close #45

## What happened 👀

- Passing data from `viewModel.coinDetail` to `CoinStatisticsSection`.
- Configure `Equatable` function of `MockCoinDetail` and `MockMarketData`.
- Add new initializers for `MockMarketData`.

## Insight 📝

- We tested the correct value for `coinDetail`, and market cap, all time high, all time low values are coinDetail's properties, so we don't need to write tests for checking market cap, all time high, all time low.

```swift
                    it("gets the correct value for coinDetail") {
                        await expect { await myCoinViewModel.coinDetail }
                            .to(equal(expectedCoins))
                    }
```

## Proof Of Work 📹

https://user-images.githubusercontent.com/25881847/211710974-0ef8d62b-4fd7-40b9-bcb3-4ac93284b622.mp4

### DIFF
--- a/CryptoPrices/Sources/MyCoin/Sources/MyCoin/CoinStatisticsSection/CoinStatisticsSection.swift
+++ b/CryptoPrices/Sources/MyCoin/Sources/MyCoin/CoinStatisticsSection/CoinStatisticsSection.swift
@@ -4,6 +4,7 @@
 //
 //  Created by Khanh on 03/01/2023.
 //
+// swiftlint:disable void_function_in_ternary
 
 import Styleguide
 import SwiftUI
@@ -14,30 +15,37 @@ struct CoinStatisticsSection: View {
         VStack {
             coinStatisticsItemView(
                 title: Strings.MyCoin.MarketCap.title,
-                price: coinDetailItem?.marketCap,
-                isPriceUp: coinDetailItem?.isMarketCapChangePercentage24HUp ?? false,
-                percentage: coinDetailItem?.marketCapChangePercentage24H
+                price: coinDetailItem.marketCap,
+                isPriceUp: coinDetailItem.isMarketCapChangePercentage24HUp,
+                percentage: coinDetailItem.marketCapChangePercentage24H
             )
+            .padding(.bottom, 24.0)
             coinStatisticsItemView(
                 title: Strings.MyCoin.Ath.title,
-                price: coinDetailItem?.ath,
-                isPriceUp: coinDetailItem?.isAthChangePercentageUp ?? false,
-                percentage: coinDetailItem?.athChangePercentage
+                price: coinDetailItem.ath,
+                isPriceUp: coinDetailItem.isAthChangePercentageUp,
+                percentage: coinDetailItem.athChangePercentage
             )
+            .padding(.bottom, 24.0)
             coinStatisticsItemView(
                 title: Strings.MyCoin.Atl.title,
-                price: coinDetailItem?.atl,
-                isPriceUp: coinDetailItem?.isAtlChangePercentageUp ?? false,
-                percentage: coinDetailItem?.atlChangePercentage
+                price: coinDetailItem.atl,
+                isPriceUp: coinDetailItem.isAtlChangePercentageUp,
+                percentage: coinDetailItem.atlChangePercentage
             )
         }
         .padding(16.0)
     }
 
-    private let coinDetailItem: CoinDetailItem?
+    private let coinDetailItem: CoinDetailItem
+    private let showData: Bool
 
-    init(coinDetailItem: CoinDetailItem?) {
+    init(
+        coinDetailItem: CoinDetailItem,
+        showData: Bool
+    ) {
         self.coinDetailItem = coinDetailItem
+        self.showData = showData
     }
 }
 
@@ -45,37 +53,39 @@ private extension CoinStatisticsSection {
 
     func coinStatisticsItemView(
         title: String,
-        price: Decimal?,
+        price: Decimal,
         isPriceUp: Bool,
-        percentage: Double?
+        percentage: Double
     ) -> some View {
         HStack {
             VStack(
                 alignment: .leading,
-                spacing: 16.0
+                spacing: 10.0
             ) {
                 Text(title)
                     .foregroundColor(Colors.textMedium.swiftUIColor)
                     .font(Fonts.Inter.medium.textStyle(.caption))
 
-                if let price {
+                ZStack(alignment: .leading) {
                     Text(price, format: .dollarCurrency)
                         .foregroundColor(Colors.titleMedium.swiftUIColor)
                         .font(Fonts.Inter.medium.textStyle(.callout))
-                } else {
+                        .opacity(showData ? 1.0 : 0.0)
+                    
                     Text(Strings.MyCoin.NoData.text)
                         .foregroundColor(Colors.titleMedium.swiftUIColor)
                         .font(Fonts.Inter.medium.textStyle(.callout))
+                        .opacity(showData ? 0.0 : 1.0)
                 }
             }
 
             Spacer()
 
-            if let percentage {
-                isPriceUp
-                ? Images.icArrowUpGreen.swiftUIImage
-                : Images.icArrowDownRed.swiftUIImage
-                
+            isPriceUp
+            ? Images.icArrowUpGreen.swiftUIImage.opacity(showData ? 1.0 : 0.0)
+            : Images.icArrowDownRed.swiftUIImage.opacity(showData ? 1.0 : 0.0)
+
+            ZStack(alignment: .trailing) {
                 Text(percentage, format: .percentage)
                     .font(Fonts.Inter.medium.textStyle(.body))
                     .foregroundColor(
@@ -83,9 +93,11 @@ private extension CoinStatisticsSection {
                         ? Colors.guppieGreen.swiftUIColor
                         : Colors.carnation.swiftUIColor
                     )
-            } else {
+                    .opacity(showData ? 1.0 : 0.0)
+                
                 Text(Strings.MyCoin.NoData.text)
                     .font(Fonts.Inter.medium.textStyle(.body))
+                    .opacity(showData ? 0.0 : 1.0)
             }
         }
     }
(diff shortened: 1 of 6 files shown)


CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
```diff
@@ -51,7 +51,7 @@ private extension MyCoinView {
             // TODO: - Remove dummy
             TimeFrameSection()
             CoinStatisticsSection(
-                coinDetailItem: CoinDetailItem(
+                coinDetailItem: viewModel.coinDetail ?? CoinDetailItem(
                     coinDetail: MockCoinDetail.single
                 )
```
minhnimble, line 56: Since we are using real data now, we can also remove the TODO, right?

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
```diff
@@ -51,7 +51,7 @@ private extension MyCoinView {
             // TODO: - Remove dummy
             TimeFrameSection()
             CoinStatisticsSection(
-                coinDetailItem: CoinDetailItem(
+                coinDetailItem: viewModel.coinDetail ?? CoinDetailItem(
```
minhnimble, line 54: Why are we falling back to the mock data if the real data is nil? This could create confusion to the user that we saw large price fluctuation here compared to real data from other sources, especially for coins with small values like dogecoin, ripple, etc. I think we should have another default data that shows all zero values or maybe hide the section completely if it is nil. What do you think? 🤔

Thieurom:
@nkhanh44 Please help rebase on the latest `develop` so we can quickly look at what changes 🙏

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
nkhanh44: Good catch 👍, the latest commit has already removed it after I rebased :D.

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
nkhanh44: Thank you for the raise 🙏 .
>I think we should have another default data that shows all zero values or maybe hide the section completely if it is nil

These are good suggestions 👍 . But I think when the real data is nil, it means that "There is no data for this". Therefore, I come up with setting it's nil data with "-". I fixed it at https://github.com/nimblehq/nimble-crypto-ios/pull/96/commits/e998377324a2ba308e06efc5cfcd5a69c11ac58f. What do you think about this approach? Thanks.

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
```diff
@@ -70,9 +70,7 @@ private extension MyCoinView {
     var coinStatisticsSection: some View {
         Section {
             CoinStatisticsSection(
-                coinDetailItem: CoinDetailItem(
-                    coinDetail: MockCoinDetail.single
-                )
+                coinDetailItem: viewModel.coinDetail
             )
         }
     }
```
Thieurom, line 76: Could you try to keep `CoinStatisticsSection` unchanged (everything is still non `Optional`, no `if let`, no `?? false`, no `noData` state ...), and just update like this for `MyCoinView`:
```swift
var coinStatisticsSection: some View {
  viewModel.coinDetail.map { coinDetail in
    Section {
      CoinStatisticsSection(coinDetailItem: coinDetail)
    }
  }
}
```

nkhanh44: Thank you for the suggestion 🙌 . It works correctly when showing an empty view if real data is nil. But I still wanted to keep the structure of the statistics section with "-" for data that are not found as the following. What do you think?

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
minhnimble: Showing the nil data as `"-"` is a good move too 👍

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
Thieurom: @nkhanh44 If that's what you expect, it's fine to go with your approach 👍 (Still to be honest, it hurts me when seeing a lot of nil coalescing (`?? false`) and imperative codes (`if let else`) in declarative SwiftUI views 😄)

One more thing, could you help to adjust some vertical spacings: between the `coinStatisticsItemView`s and the inner of the `coinStatisticsItemView` itself 🙏

Simulator Screen Shot - iPhone 14 - 2023-01-12 at 14 56 05

From the screenshot of our app, the (inner) spacing between `"Market Cap"` and `"$168,926,..."` is bigger than that of between `"$168,926,..."` and `"All Time High"` (outer one). So visually it's hard to see the clear separation between those groups, it even can make us confused about which group `"$168,926,..."` should it belong to.

The design looks good though:
Screenshot 2023-01-12 at 14 56 24

nkhanh44: @Thieurom Thank you for the raise, I agree with your concern 👍. I refactored the code in the coin's statistic without a lot of nil coalescing (?? false) and imperative codes (if let else). Please help me to review if it's fine with you 🙏 .

CryptoPrices/Sources/MyCoin/Sources/MyCoin/CoinStatisticsSection/CoinStatisticsSection.swift
```diff
@@ -52,30 +60,45 @@ private extension CoinStatisticsSection {
...
+                        isPriceUp
+                        ? Colors.guppieGreen.swiftUIColor
+                        : Colors.carnation.swiftUIColor
+                    )
+                    .opacity(showData ? 1.0 : 0.0)
+                
+                Text(Strings.MyCoin.NoData.text)
+                    .font(Fonts.Inter.medium.textStyle(.body))
+                    .opacity(showData ? 0.0 : 1.0)
+            }
```
Thieurom, line 101: I don't know whether I like this change or the previous though 😄
Because we replace:
```swift
if let something = something {
  ViewA()
} else {
  ViewB()
}
```
with
```swift
ZStack {
  ViewA()
  .opacity(condition ? 1.0 : 0.0)

  ViewB()
  .opacity(condition ? 0.0 : 1.0)
}
```
along with introducing some _ternary conditional_ operators (it's basically the syntactic sugar for `if ... else` I believe)

Anyway, I think it's just a matter of preference. I really appreciate your effort striking to make the code cleaner, easy to read, and maintainable. And we have many chances to tackle this more seriously later. So let's go 🚀

Thieurom (approved):

CryptoPrices/Sources/MyCoin/Sources/MyCoin/CoinStatisticsSection/CoinStatisticsSection.swift
```diff
@@ -4,6 +4,7 @@
 //
 //  Created by Khanh on 03/01/2023.
 //
+// swiftlint:disable void_function_in_ternary
```
minhnimble, line 7: Can we fix this by turning the `isPriceUp` ternary check into an `if else` check? I think we should try to fix the raised swiftlint issues rather than ignoring them unless there is no way to resolve them with our convention 🙏

minhnimble (commented):
Just a few small concerns left, the rest is really nice work! 🚀

CryptoPrices/Sources/MyCoin/Sources/MyCoin/CoinStatisticsSection/CoinStatisticsSection.swift
nkhanh44: Thank you so much 🙏. I fixed it at https://github.com/nimblehq/nimble-crypto-ios/pull/96/commits/a000bbe862a1a053b222617697dfc41585b1a87c.

minhnimble (approved):
🚀

PR #98, opened 2023-01-11
[#63] [Integrate] As a user, I can pull to refresh data on the Detail screen

### LINKED ISSUE #63: [Integrate] As a user, I can pull to refresh data on the Detail screen

nkhanh44:
- Close #63

## What happened 👀

- Add refreshable for calling all APIs when pulling to refresh.

## Insight 📝
- The loading indicator's put on the same as the design. Because when applying the `refreshable` modifier to the `List`, the loading indicator will place on the top of the `List` and below the navigation bar as the default.

## Proof Of Work 📹

https://user-images.githubusercontent.com/25881847/212016035-4bb1a52a-45e6-4782-b4e9-543bc6a6119e.mp4

### DIFF
--- a/CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
+++ b/CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
@@ -48,6 +48,9 @@ public struct MyCoinView: View {
                 await viewModel.fetchChartPricesData(id: myCoinState.id, timeFrameItem: selectedTimeFrameItem)
             }
         }
+        .refreshable {
+            await viewModel.fetchData(id: myCoinState.id, timeFrameItem: .init(timeFrame: .oneDay))
+        }
     }
 
     @ObservedObject private var viewModel: MyCoinViewModel
--- a/CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
+++ b/CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
@@ -15,6 +15,7 @@ public struct MyCoinView: View {
     @EnvironmentObject var myCoinState: MyCoinState
 
     @State private var selectedTimeFrameItem: TimeFrameItem = .init(timeFrame: .oneDay)
+    @State private var tempSelectedTimeFrameItem: TimeFrameItem = .init(timeFrame: .oneDay)
 
     public var body: some View {
         List {
@@ -48,8 +49,16 @@ public struct MyCoinView: View {
                 await viewModel.fetchChartPricesData(id: myCoinState.id, timeFrameItem: selectedTimeFrameItem)
             }
         }
+        .onReceive(viewModel.$isSuccess) { isSuccess in
+            if isSuccess {
+                tempSelectedTimeFrameItem = selectedTimeFrameItem
+            } else {
+                selectedTimeFrameItem = tempSelectedTimeFrameItem
+            }
+        }
         .refreshable {
-            await viewModel.fetchData(id: myCoinState.id, timeFrameItem: .init(timeFrame: .oneDay))
+            selectedTimeFrameItem = .init(timeFrame: .oneDay)
+            await viewModel.fetchCoinDetail(id: myCoinState.id)
         }
     }
 
--- a/CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinViewModel.swift
+++ b/CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinViewModel.swift
@@ -15,6 +15,7 @@ import UseCaseProtocol
 
     @Published public private(set) var coinDetail: CoinDetailItem?
     @Published public private(set) var chartData: [ChartDataPointUIModel] = []
+    @Published public private(set) var isSuccess = false
 
     public init(
         coinDetailUseCase: CoinDetailUseCaseProtocol,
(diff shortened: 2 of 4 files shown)


Thieurom (approved):
🚀

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
```diff
@@ -47,6 +47,9 @@ public struct MyCoinView: View {
                 await viewModel.fetchChartPricesData(id: myCoinState.id, timeFrameItem: selectedTimeFrameItem)
             }
         }
+        .refreshable {
+            await viewModel.fetchData(id: myCoinState.id, timeFrameItem: .init(timeFrame: .oneDay))
```
minhnimble, line 51: It looks like we forget to update the UI of the selected filter when refreshing the chart data to `1D` filter. For example, while I am using the app, I switch the filter to `1M` and then trigger a refresh, it doesn't select back to the `1D` filter:

https://user-images.githubusercontent.com/70877098/212587847-691c44a6-691e-432c-aac6-c7cbcbbb40af.mov

And in case the refresh was unsuccessful, we might need to keep showing the same data and keep the same filter also. Please help check 🙏

minhnimble (commented):
There is a small behavior issue, the rest LGTM as it is a simple logic update 🚀

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
nkhanh44: Good catch 🙏 . I forgot that we detected the change of `selectedTimeFilterItem`, so call `fetchData` there is unnecessary.
>And in case the refresh is unsuccessful, we might need to keep showing the same data and keep the same filter also. Please help check 🙏

Thank you for raising another edge case. Good point. I fixed it at  https://github.com/nimblehq/nimble-crypto-ios/pull/98/commits/444e39b98ac45e1bbaaecdaece67099f492a779e :D.

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinViewModel.swift
```diff
@@ -15,6 +15,7 @@ import UseCaseProtocol
 
     @Published public private(set) var coinDetail: CoinDetailItem?
     @Published public private(set) var chartData: [ChartDataPointUIModel] = []
+    @Published public private(set) var isSuccess = false
```
minhnimble, line 18: Please make the flag name more specific for better clarification 🙏
```suggestion
    @Published public private(set) var isCoinDataRefreshedSuccessfully = false
```

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
```diff
@@ -48,6 +49,16 @@ public struct MyCoinView: View {
...
         }
+        .onReceive(viewModel.$isSuccess) { isSuccess in
+            if isSuccess {
+                tempSelectedTimeFrameItem = selectedTimeFrameItem
+            } else {
+                selectedTimeFrameItem = tempSelectedTimeFrameItem
+            }
+        }
+        .refreshable {
+            selectedTimeFrameItem = .init(timeFrame: .oneDay)
```
minhnimble, line 60: Oh so now when we refresh the `Coin Details` screen, we will only refresh the chart region data with 1D filter but we don't trigger to reload the `coinDetail` data anymore? Is this something expected? 🤔

nkhanh44: My bad 😢 . I added the `fetchCoinDetail` into `refreshable` modifier at https://github.com/nimblehq/nimble-crypto-ios/pull/98/commits/de4774a78c84143af07df2964e1b1ac2e9701ff1.

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinViewModel.swift
nkhanh44: >And in case the refresh is unsuccessful, we might need to keep showing the same data and keep the same filter also.

As we handle the refresh in unsuccessful requests, I think we should handle it too when we switch among timeframes in the following video:
https://user-images.githubusercontent.com/25881847/212600288-d60ec3ee-1af8-4f81-9e0e-67d0749895c4.mp4
What do you think? 🤔

minhnimble: > I think we should handle it too when we switch among timeframes in the following video

That is definitely a good point @nkhanh44 👍 Now the need for having a popup message arises critically as the user might get confused why they can't switch the filter, but yeah I think we can uniformly handle the error popups in a separated PR 😁

minhnimble (approved):
LGTM 🚀

PR #86, opened 2023-01-04
[#41] [UI] As a user, I can see a timeframe bar on the Detail screen

nkhanh44:
- Close #41 

## What happened 👀

- Created UI Timeframe in Detail Screen. 

## Insight 📝

- Whenever we select a timeframe item, we will compare its `TimeFrameItem` to figure out which item should be highlighted.

## Proof Of Work 📹

Dark mode:

https://user-images.githubusercontent.com/25881847/210498972-74edde7b-ff6b-413f-887e-98531c815d38.mp4

Light mode:

https://user-images.githubusercontent.com/25881847/210499005-61f15499-23b0-4724-811a-1fcf0a28040f.mp4

PR #99, opened 2023-01-12
[#43] [Integrate] As a user, I can see timeframe bar on the Detail screen

### LINKED ISSUE #43: [Integrate] As a user, I can see timeframe bar on the Detail screen
## Why
- The user should be able to see the timeframe bar on the Detail screen.

## Acceptance Criteria

- On the Detail screen, the following timeframe bar can be seen:

_in the light mode:_
Screenshot 2022-12-12 at 15 54 45

_in the dark mode:_
Screenshot 2022-12-12 at 15 55 15

- `1D` is the default selection when navigating to the detail screen.
- Whenever a bar item `1D, 7D, 1M, 1Y, or 5Y` is selected, it will be highlighted with `#00BDB0`. And the corresponding chart will be updated with API response from #42 .

```JSON
{
  "prices": [
    [
      1669284289704,
      16595.246079387816
    ],
    [
      1669284585059,
      16602.076684659474
    ],
    [
      1669284921814,
      16606.11839794301
    ],
    [
      1669285188016,
      16606.20441243831
}
```

Ethereum

nkhanh44:
- Close #43

## What happened 👀

- Added filter parameter to `fetchChartPricesData` function to request data chart based on timeframe (1D, 1W, 1M, 1Y, 5Ys).
- Added animation for the timeframe bar.
- Updated Unit test in `MyCoinViewModelSpec`.

## Insight 📝

- Changed `List` to `ScrollView`, because when using `List`, we have a problem with the timeframe bar as the following:

https://user-images.githubusercontent.com/25881847/212232444-ee9de8c7-6fd0-4fa7-bdd9-062bd8e7f393.mp4

And for the concern of refreshing, I will handle it at [this PR](https://github.com/nimblehq/nimble-crypto-ios/pull/98).

- For [this suggestion](https://github.com/nimblehq/nimble-crypto-ios/pull/86#discussion_r1063071065) from anh Thieu, I tried to implement the timeframe bar with the approach of the built-in `Picker`, but it's not closed to the design. Therefore, I will skip this attempt and still keep this current implementation.
## Proof Of Work 📹

https://user-images.githubusercontent.com/25881847/212232594-8454a2f6-e051-4a6a-a954-b82a8c2a7cb3.mp4

### DIFF
--- a/CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
+++ b/CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
@@ -13,9 +13,10 @@ private typealias Section = Styleguide.Section
 public struct MyCoinView: View {
 
     @EnvironmentObject var myCoinState: MyCoinState
+    @State private var selectedTimeFrameItem: TimeFrameItem = .init(timeFrame: .oneDay)
 
     public var body: some View {
-        List {
+        ScrollView {
             currentPriceSection
             priceChartSection
             coinStatisticsSection
@@ -39,7 +40,12 @@ public struct MyCoinView: View {
             footerView
         })
         .task {
-            await viewModel.fetchData(id: myCoinState.id)
+            await viewModel.fetchData(id: myCoinState.id, timeFrameItem: selectedTimeFrameItem)
+        }
+        .onChange(of: selectedTimeFrameItem) { _ in
+            Task {
+                await viewModel.fetchData(id: myCoinState.id, timeFrameItem: selectedTimeFrameItem)
+            }
         }
     }
 
@@ -63,7 +69,7 @@ private extension MyCoinView {
             PriceLineChartSection(viewModel.chartData)
                 .frame(height: 196.0)
 
-            TimeFrameSection()
+            TimeFrameSection(selected: $selectedTimeFrameItem)
         }
     }
 
--- a/CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
+++ b/CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
@@ -44,7 +44,7 @@ public struct MyCoinView: View {
         }
         .onChange(of: selectedTimeFrameItem) { _ in
             Task {
-                await viewModel.fetchData(id: myCoinState.id, timeFrameItem: selectedTimeFrameItem)
+                await viewModel.fetchChartPricesData(id: myCoinState.id, timeFrameItem: selectedTimeFrameItem)
             }
         }
     }
--- a/CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
+++ b/CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
@@ -13,6 +13,7 @@ private typealias Section = Styleguide.Section
 public struct MyCoinView: View {
 
     @EnvironmentObject var myCoinState: MyCoinState
+
     @State private var selectedTimeFrameItem: TimeFrameItem = .init(timeFrame: .oneDay)
 
     public var body: some View {
--- a/CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
+++ b/CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
@@ -16,7 +16,7 @@ public struct MyCoinView: View {
     @State private var selectedTimeFrameItem: TimeFrameItem = .init(timeFrame: .oneDay)
 
     public var body: some View {
-        ScrollView {
+        List {
             currentPriceSection
             priceChartSection
             coinStatisticsSection
--- a/CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinViewModel.swift
+++ b/CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinViewModel.swift
@@ -33,20 +33,19 @@ import UseCaseProtocol
         }
     }
 
-    // TODO: Support different filters
-    public func fetchChartPricesData(id: String) async {
+    public func fetchChartPricesData(id: String, timeFrameItem: TimeFrameItem) async {
         do {
-            let dataPoints = try await getChartPricesUseCase.execute(coinID: id, filter: .oneDay)
+            let dataPoints = try await getChartPricesUseCase.execute(coinID: id, filter: timeFrameItem.timeFrame)
             chartData = dataPoints.map { ChartDataPointUIModel(dataPoint: $0) }
         } catch {
             // TODO: Handle errors
         }
     }
 
-    public func fetchData(id: String) async {
+    public func fetchData(id: String, timeFrameItem: TimeFrameItem) async {
         await withTaskGroup(of: Void.self) { taskGroup in
             taskGroup.addTask { await self.fetchCoinDetail(id: id) }
-            taskGroup.addTask { await self.fetchChartPricesData(id: id) }
+            taskGroup.addTask { await self.fetchChartPricesData(id: id, timeFrameItem: timeFrameItem) }
         }
     }
 }
--- a/CryptoPrices/Sources/MyCoin/Sources/MyCoin/TimeFrameSectionView/TimeFrameSection.swift
+++ b/CryptoPrices/Sources/MyCoin/Sources/MyCoin/TimeFrameSectionView/TimeFrameSection.swift
@@ -45,8 +45,17 @@ struct TimeFrameSection: View {
             .padding(.horizontal, 40.0)
             .padding(.vertical, 24.0)
         }
-        .onChange(of: selected) { newValue in
-            print(newValue)
+        .animation(.default, value: selected)
+    }
+}
+
+#if DEBUG
+struct TimeFrameSection_Previews: PreviewProvider {
+
+    static var previews: some View {
+        Preview {
+            TimeFrameSection(selected: .constant(.init(timeFrame: .oneDay)))
         }
     }
 }
+#endif
--- a/CryptoPrices/Sources/MyCoin/Sources/MyCoin/TimeFrameSectionView/TimeFrameSection.swift
+++ b/CryptoPrices/Sources/MyCoin/Sources/MyCoin/TimeFrameSectionView/TimeFrameSection.swift
@@ -10,7 +10,7 @@ import SwiftUI
 
 struct TimeFrameSection: View {
     
-    @State private var selected: TimeFrameItem = .init(timeFrame: .oneDay)
+    @Binding var selected: TimeFrameItem
     
     private let timeFrameList: [TimeFrameItem] = [
         .init(timeFrame: .oneDay),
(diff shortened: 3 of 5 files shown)


CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
```diff
@@ -39,7 +40,12 @@ public struct MyCoinView: View {
             footerView
         })
         .task {
-            await viewModel.fetchData(id: myCoinState.id)
+            await viewModel.fetchData(id: myCoinState.id, timeFrameItem: selectedTimeFrameItem)
+        }
+        .onChange(of: selectedTimeFrameItem) { _ in
+            Task {
+                await viewModel.fetchData(id: myCoinState.id, timeFrameItem: selectedTimeFrameItem)
```
minhnimble, line 47: Wait, I don't think we should call `fetchData` here, because it already reloads the coinDetail information from the server again, but that is not necessary when we change the filter. As we just want to reload the chart only, right? 🤔

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinViewModel.swift
```diff
@@ -33,20 +33,19 @@ import UseCaseProtocol
         }
     }
 
-    // TODO: Support different filters
-    public func fetchChartPricesData(id: String) async {
+    public func fetchChartPricesData(id: String, timeFrameItem: TimeFrameItem) async {
         do {
-            let dataPoints = try await getChartPricesUseCase.execute(coinID: id, filter: .oneDay)
+            let dataPoints = try await getChartPricesUseCase.execute(coinID: id, filter: timeFrameItem.timeFrame)
             chartData = dataPoints.map { ChartDataPointUIModel(dataPoint: $0) }
```
minhnimble, line 39: Suggestion: Since 5 different filters will introduce a lot of chart data points, it feels inefficient that we just reload the chart every time from the server and discard the previous chart filter data. 

=> We should apply some caching mechanism here and only load the chart from the server the first time selecting a filter so that when the user switches the filters fast, we won't spam the requests to the server and improve the rendering time in a session. We will also have the capability to refresh later on so we can clean up the data at that time. What do you think? 🙏

FYI, here are some options to cache the chart data for us to consider. Let me know your thought 🙏:
  - Create a dictionary with the key type is `TimeFrameItem` and the value type is `[ChartDataPointUIModel]`.  This will allow storing the data for all filters directly in this `MyCoinViewModel` as an in-memory option.
  - Set up a local database like `Realm` or `CoreData` to store/retrieve the chart data accordingly. This way we can keep the cache data alive across sessions but required some forefront setups. This can be done in a separate PR if we want to follow this approach.

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
nkhanh44: Good catch 🙌 . I will fix it.

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinViewModel.swift
nkhanh44: Thank you for the suggestion 🙏 . It's a good point to apply some caching mechanisms to avoid spamming requests the server 👍, especially when the user quickly changes the timeframe for any purpose. But in crypto, the data changes so frequently (I mean in a second), caching the data is not quite the best practice. I also checked the timeframe bar on the [coingecko](https://www.coingecko.com/), they do send the request every time we change the timeframe, even though the chart shows not much difference :D. Therefore, I think we can keep this current implementation, even though our app only uses the Free API for not real-time data 🙏 . What do you think?

> - Create a dictionary with the key type is TimeFrameItem and the value type is [ChartDataPointUIModel]. This will allow storing the data for all filters directly in this MyCoinViewModel as an in-memory option.
> - Set up a local database like Realm or CoreData to store/retrieve the chart data accordingly. This way we can keep the cache data alive across sessions but required some forefront setups. This can be done in a separate PR if we want to follow this approach.

Again, thank you so much for the suggestion along with the solutions 👍 . Great suggestion.

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
nkhanh44: I fixed it at https://github.com/nimblehq/nimble-crypto-ios/pull/99/commits/08a5951398f368ea09c46de2228e482563272bdf 🙏 .

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinViewModel.swift
minhnimble: Ohhh, I don't think we are displaying a real-time chart and price here or anything, so that is why I suggest caching and I believe it would not be a bad call. But yeah, this is in the end a suggestion and we can have different views about this.

In fact, caching is indeed more of a complex approach as there can be extra things that need to be decided further down the road such as managing the expiration of the cache, auto-refreshing cache data, selecting cache type (in-memory, local database), etc. Thus, reloading the chart after every filter change is a much simpler implementation logic to deal with 👍

-> Let's keep the current implementation for now until we found it extra annoying when it reaches the API limits too often 🙏

nkhanh44: >In fact, caching is indeed more of a complex approach as there can be extra things that need to be decided further down the road such as managing the expiration of the cache, auto-refreshing cache data, selecting cache type (in-memory, local database), etc.

Nice 👍 .

Thieurom: As @nkhanh44 mentioned, usually, for time-sensitive applications like stock market price, voting, etc, the data is updated so frequently so caching does not mean much IMHO. However, I think even for that kind of application, we won't update the data on the user's demand (like switching to another time filter or pulling to refresh) or time interval by making multiple network requests (polling). Instead (at least in my experience) the app maintains a single connection with the server through WebSocket 🙏 

It's still possible to implement effective caching but we would do that at the lower layer like the network. It's because the network layer can access the `HTTPResponse` which contains the information about data invalidation like `Cache-Control`, `Last-Modified`, ...

But as @minhnimble mentioned, we shouldn't bother with this complexity much at this point, so keeping the current implementation is fine for me too 🙏

CryptoPrices/Sources/MyCoin/Sources/MyCoin/TimeFrameSectionView/TimeFrameSection.swift
```diff
@@ -10,7 +10,7 @@ import SwiftUI
 
 struct TimeFrameSection: View {
     
-    @State private var selected: TimeFrameItem = .init(timeFrame: .oneDay)
+    @Binding var selected: TimeFrameItem
```
Thieurom, line 13: One thing I just noticed, we actually can use `TimeFilter` directly here, there's no need for the `TimeFrameItem` to exist, which wraps `timeFrame: TimeFilter` without anything else. But it's fine to keep this as is though 🙏

CryptoPrices/Sources/MyCoin/Sources/MyCoin/TimeFrameSectionView/TimeFrameSection.swift
```diff
@@ -25,7 +25,6 @@ struct TimeFrameSection: View {
             HStack {
                 ForEach(timeFrameList) { item in
                     Button(item.title) {
```
Thieurom, line 27: Feel free to ignore this 🙏 We can omit the conformance to `Identifiable` for the `TimeFrameItem`, i.e remove `let id = UUID()` by:
```swift
ForEach(timeFrameList, id: \.timeFrame) { item in
  // ...
}
```

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
```diff
@@ -13,9 +13,10 @@ private typealias Section = Styleguide.Section
 public struct MyCoinView: View {
 
     @EnvironmentObject var myCoinState: MyCoinState
+    @State private var selectedTimeFrameItem: TimeFrameItem = .init(timeFrame: .oneDay)
 
     public var body: some View {
-        List {
+        ScrollView {
```
Thieurom, line 19: I don't know if this is Apple's intention or a bug, but it seems that by default the `List` _observes_ or _handles_ tapping views inside of it. So when our view, which is the `TimeFrameSection` has a button with action, it gets confused when we tap.

Try to keep this using `List`, but in the `TimeFrameSection` add this line:
```swift
ForEach(timeFrameList) { item in
  Button(item.title) {
    selected = TimeFrameItem(timeFrame: item.timeFrame)
  }
  .buttonStyle(.borderless) 👈  add this
  // ...
}
```

CryptoPrices/Sources/MyCoin/Sources/MyCoin/TimeFrameSectionView/TimeFrameSection.swift
nkhanh44: Thanks for the suggestion 🙏

CryptoPrices/Sources/MyCoin/Sources/MyCoin/TimeFrameSectionView/TimeFrameSection.swift
nkhanh44: Thanks, I fixed it at https://github.com/nimblehq/nimble-crypto-ios/pull/99/commits/2fab7d69ff3151a83b4dd512405f946144981c0d

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
nkhanh44: Yeah, the behavior's so weird :D. The solution is excellent 🙏 . I fixed it at https://github.com/nimblehq/nimble-crypto-ios/pull/99/commits/2fab7d69ff3151a83b4dd512405f946144981c0d. Thank you.

Thieurom (approved):
🚀

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
```diff
@@ -13,6 +13,7 @@ private typealias Section = Styleguide.Section
 public struct MyCoinView: View {
 
     @EnvironmentObject var myCoinState: MyCoinState
+    @State private var selectedTimeFrameItem: TimeFrameItem = .init(timeFrame: .oneDay)
```
minhnimble, line 16: Optional, maybe adding a line break above for grouping different type of var, just a code presentation improvement 😁

minhnimble (approved):
The rest LGTM, naiseee 🚀

CryptoPrices/Sources/MyCoin/Sources/MyCoin/MyCoinView/MyCoinView.swift
nkhanh44: 🙏 I fixed it at https://github.com/nimblehq/nimble-crypto-ios/pull/99/commits/1790e7a7915af1aece3b9751017a8ee0f4628d48

PR #106, opened 2023-01-18
[#103] [UI+Integrate] Display a loading spinner while fetching the data

### LINKED ISSUE #103: [UI+Integrate] Display a loading spinner while fetching the data
## Why

When the app first launches or navigates to a new screen, there's no indication that the app is fetching some data from the network. A loading spinner will add that missing part to improve the user experience.

The loading spinner also helps when the network request fails. Since the app has not implemented error handling, a (stopped) loading spinner will hint to the user that the network request has been finished.

- Show the loading spinner upon the `Home` and `MyCoin` appearing on the app.
- The loading spinner won't be shown when the user pulls to refresh.

## Who Benefits?

Users and developers.

Thieurom:
- Close #103

## What happened 👀

- Create the `CircularProgressView`, which shows a (static) ring view with a given progress value.
- Create the `spinner` view modifier, which can be attached to a view to show a `CircularProgressView` instance with a rotation animation. Apply the modifier to `HomeView` and `MyCoinView`.
- Update the method calls in `HomeViewModel` and `MyCoinViewModel` with a `isRefreshing` parameter to distinguish the context (view appearing or data refreshing)
- Update the lifetime of `Factory`'s creation dependencies:
  - It started with a bug. When users tap a coin item on the **Home** to navigate to the **Detail**, the view displays the data from the previous state while waiting for the network response. At first, I thought there could be something wrong with the `@EnvironmentObject.` Actually, the root cause is that we pass the same instance of `MyCoinViewModel` when we create the new `MyCoinView`:
    ```swift
    struct AppView: View {
      @Injected(Container.myCoinViewModel) private var myCoinViewModel
      
      var body: some View {
        // ...  
        MyCoinView(viewModel: myCoinViewModel)
        // ...
      }
    }
    ```
    hence the update is passing the instance from the `Container` instead of keeping around an injected one:
    ```swift
    MyCoinView(viewModel: Container.myCoinViewModel())
    ```
  - From this, I reconsidered the lifetime of the dependencies:

    For the service types like the network, the `Repository`s, and the `UseCase`s, it's fine to keep only one instance of them across the app since they contain just some kind of work on demand (fetching and returning data). They don't keep any internal state, so there is no worry about conflict or race conditions. The `Factory` provides some predefined scope (we can define a custom one) like `cached`, `singleton`, or `shared` ... For our simple app, there's seem no difference between these so I went with `cached`.

    For the presentation layer types like view models, in general, their lifetime should be the same as their corresponding views, except for the `HomeView` since it is the root view of the app (it's always there). So the `HomeViewModel` can be cached.
  

## Insight 📝

While creating the loading spinner, at first, I styled it with the same purple color as [the Android did](https://github.com/nimblehq/jetpack-compose-crypto) (the color choice seems to be related to the MaterialDesign thing)

IMHO, the purple color looks out of place with others from the existing palette, especially in light mode. It should be a complement or a contrast one, the purple is not one of these (I'm no expert in UI/UX design 🙏) Here the app shows the spinner with the purple color in case you wonder:
| Dark mode | Light mode |
|---|---|
|<video src="https://user-images.githubusercontent.com/17972674/213356008-eacaf011-f580-4906-be2e-4e6e825e1d66.mp4">|<video src="https://user-images.githubusercontent.com/17972674/213356013-5e5f4f7d-f3a0-4b49-98a9-62704ebcdb27.mp4">|

So I tried to pick one from the existing colors and found that `tiffanyBlue` looks good in light and dark modes.

But it's just a matter of preferences. Please raise your thoughts, and I'm happy to make another choice 🙏 

## Proof Of Work 📹

| Dark mode | Light mode |
|---|---|
|<video src="https://user-images.githubusercontent.com/17972674/213354848-0a4fa49c-fb76-429a-8876-a711fd8bb07b.mp4">|<video src="https://user-images.githubusercontent.com/17972674/213354851-b4c0c417-229a-432e-9803-bd61cb9020b0.mp4">|


CryptoPrices/Sources/Styleguide/Sources/Styleguide/CommonViews/CircularProgressView.swift
```diff
@@ -0,0 +1,26 @@
...
+    private let tint: Color
+
+    public var body: some View {
+        Circle()
+            .trim(from: 0.0, to: progress)
+            .stroke(tint, lineWidth: 4.0)
+            .frame(width: 48.0, height: 48.0)
+    }
+
+    public init(progress: Double, tint: Color = .gray) {
```
nkhanh44, line 22: It seems we don't use the default (.gray) color anywhere. Do you think we should replace the default color `.gray` with `Colors.tiffanyBlue.swiftUIColor`?

nkhanh44 (approved):
Great work with the spinner customization and the lifetime of the dependencies 👍. Also, `tiffanyBlue` is good to me. Purple is not correlated to the UI app at all, IMO 😄.

minhnimble (approved):
Like @nkhanh44's comment, I also prefer `tiffanyBlue` for a more consistent theme color option than `purple`. Apart from Khanh's comment, rest LGTM 🚀

Thieurom:
@minhnimble @nkhanh44 Thank you guys 🙏 Updated with https://github.com/nimblehq/nimble-crypto-ios/pull/106/commits/d7bc9f4aa62bb413c085a37a521245a8a01954b2

